PR #2, opened 2024-07-15
Feat: Relax checks for restoring a sample

voegtlel:
Relaxes checks for restoring a sample.

* Allow restoring a batch of samples through `BatchDataset` (typing is not completely correct here, because nesting applies)
* Allow restoring samples from a non-stateless `IterMapDataset` and `MapDataset`. Purpose of this allowance is to enable restoring when this is actually not set properly (e.g. via Metadataset)

The use-case for this is to allow random access for dataset-comparisons for stateless `TaskEncoder` (i.e. not depending on a random state). The user may store all `__restore_key__`s of samples, and restore those samples (effectively by-index).

Example Usage:
```py
def print_error_handler(exc: Exception, key: Optional[str]):
    traceback.print_exc()

dataset = get_val_dataset(
    dataset_path,
    worker_config=...,
    batch_size=1,
    task_encoder=...,
    handler=print_error_handler,
)

index = {}
for data in dataset:
    for key, restore_key in zip(data['__keys__'], data['__restore_keys__']):
        index[key] = restore_key

# Use the index to random access a sample. This is not fast, but easy.
for key, restore_key in sorted(index.items()):
    restored_sample = dataset.restore_sample([restore_key])
```

### DIFF
--- a/src/megatron/energon/wrappers/batch_dataset.py
+++ b/src/megatron/energon/wrappers/batch_dataset.py
@@ -112,12 +112,28 @@ class BatchDataset(
                 self.error_handler(e, batch)
 
     def can_restore_sample(self) -> bool:
-        return False
+        # Cannot really verify if the returned elements contain a __restore_key__.
+        # If the user wants to use this, well...
+        return self.dataset.can_restore_sample()
 
-    def restore_sample(self, index: Tuple[Union[str, int], ...]) -> T_batch:
-        # TODO: We'd need to store multiple indices to restore a batch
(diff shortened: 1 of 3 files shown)


philipp-fischer:
This feature to always allow restoring samples of BatchDataset and MapDataset so far has only one use case. It is needed to fetch individual samples by their key for dataset comparison.
On the other hand it adds additional complexity to the restore code, hence suggest to put this PR on hold.

voegtlel:
Closed in favor of #12

PR #3, opened 2024-07-19
Fix epath with relative usage

voegtlel:
Allows epath with s3 as target and relative path usage (i.e. ".."). S3 does not support URLs containing "..", thus this must be resolved to the respective parent.

### DIFF
--- a/src/megatron/energon/epathlib/epath.py
+++ b/src/megatron/energon/epathlib/epath.py
@@ -63,7 +63,8 @@ class EPath:
                     # rclone paths are treated as absolute
                     # The first part of the path will be the rclone remote name
                     path = "/" + path
-                self.internal_path = PurePosixPath(path)
+
+                self.internal_path = self._resolve(path)
 
                 remote_name = self.internal_path.parts[1]
                 self.fs, self.s3_args = self.create_s3fs_from_rclone_remote(
@@ -73,6 +74,28 @@ class EPath:
                 self.num_fs_path_parts = 2  # Root and remote name
             else:
                 raise ValueError(f"Unknown protocol: {protocol}")
+
+    @staticmethod
+    def _resolve(path: Union[str, PurePosixPath]) -> PurePosixPath:
+        """Resolve a path, removing .. and . components."""
+        if isinstance(path, str):
+            path = PurePosixPath(path)
+        parts = path.parts
+        if parts[0] != "/":
+            raise ValueError("Only absolute paths are supported")
+        if ".." in parts or "." in parts:
+            new_parts = []
+            for part in parts[1:]:
+                if part == "..":
+                    if len(new_parts) == 0:
+                        raise ValueError(f"Path above root: {path}")
+                    new_parts.pop()
+                elif path == ".":
+                    pass
+                else:
+                    new_parts.append(part)
+            path = PurePosixPath("/", *new_parts)
+        return path
 
     @staticmethod
     def _split_protocol(path):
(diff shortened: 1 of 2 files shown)


src/megatron/energon/epathlib/epath.py
```diff
@@ -74,6 +75,24 @@ def __init__(
...
+            path = PurePosixPath(path)
+        parts = path.parts
+        if parts[0] != "/":
+            raise ValueError("Only absolute paths are supported")
+        if ".." in parts or '.' in parts:
+            new_parts = []
+            for part in parts:
+                if part == '..':
+                    new_parts.pop()
+                else:
```
philipp-fischer, line 90: Since we are expecting `.` parts in the path (see `if` above), suggest to add an `elif` branch to handle `.` parts and remove them.

src/megatron/energon/epathlib/epath.py
```diff
@@ -74,6 +75,24 @@ def __init__(
...
+                    new_parts.append(part)
+            parts = new_parts
+            path = PurePosixPath("/", *parts)
```
philipp-fischer, line 93: Above we are checking that the first part is `/` (absolute path) and here we are adding another `/`?

tests/test_epathlib.py
```diff
@@ -90,6 +88,31 @@ def test_glob(self):
         p = EPath("/tmp").glob("epathtestfile_[0-3].bin")
         assert len(list(p)) == 4
 
+    def test_s3_path_resolution(self):
```
philipp-fischer, line 91: Is this test working? Earlier, I had to remove all tests that rely on S3 connectivity. See here: https://github.com/NVIDIA/Megatron-Energon/commit/1be5fb7881f65ebefb441763a3b337f66297b338

src/megatron/energon/epathlib/epath.py
```diff
@@ -74,6 +75,24 @@ def __init__(
             else:
                 raise ValueError(f"Unknown protocol: {protocol}")
 
+    @staticmethod
+    def _resolve(path: Union[str, PurePosixPath]) -> PurePosixPath:
```
philipp-fischer, line 79: Can you add a docstring? I guess all that this does is getting rid of `..` and `.`?

philipp-fischer (changes requested):

tests/test_epathlib.py
voegtlel: Yes, it doesn't use S3 connectivity really, but is just checking the path resolution

src/megatron/energon/epathlib/epath.py
voegtlel: Done

src/megatron/energon/epathlib/epath.py
voegtlel: Right, added it

src/megatron/energon/epathlib/epath.py
voegtlel: Fixed. Strange that it worked nevertheless

voegtlel (commented):
Applied comments

tests/test_epathlib.py
philipp-fischer: Ok. My plan was to mock the S3 connection to bring back the tests (when we have time).

src/megatron/energon/epathlib/epath.py
```diff
@@ -54,7 +54,7 @@ def __init__(
         elif isinstance(initial_path, str):
             protocol, path = self._split_protocol(initial_path)
             if protocol is None:
-                self.internal_path = PurePosixPath(path)
+                self.internal_path = PurePosixPath(PathlibPath(path).absolute())
```
philipp-fischer, line 57: I debugged why the tests fail and this turns out to be the problem. This line makes all EPaths absolute, but this is wrong. 
We do need relative EPaths. Those should be allowed as long as it's not an rclone path.

I haven't checked the implications of removing `.absolute()` here. I guess you put it for a reason so it might break other things. Let's discuss this.

philipp-fischer (changes requested):

src/megatron/energon/epathlib/epath.py
voegtlel: Oh sorry, my push with that fixed failed. Reverted that change now.

philipp-fischer (approved):

PR #5, opened 2024-08-01
Slicing batch feature and refactoring for better restore key support

voegtlel:
## MINOR BREAKING CHANGES
* Randomness will change in the task encoder if using `@stateless(restore_seeds=True)` for augmentations etc.
* Dataset checkpoint have additional fields / types for storing the sample index, thus are not compatible to (loading) previous checkpoints.

## New features
* **Main feature**: Implement new `TaskEncoder.slice_batch`, which efficiently allows for slicing up batches from a buffer, and can be saved/restored
* Implement the restore key through all functions, if marked with `@stateless` in the task encoder (and giving an option to .
* Allow restoring a sample from outside of the data loader. This is a nice side effect of the refactoring, allowing to debug the task encoder reproducibly.
* All use functions can access a sample index now (via the `TaskEncoder.current_sample_index`, or directly via `WorkerConfig.active_worker_sample_index` and respectively `TaskEncoder.current_batch_index` / `WorkerConfig.active_worker_batch_index`)

Supersedes #2


docs/source/packing_slicing_batching.md
```diff
@@ -0,0 +1,56 @@
...
+    properly sized input samples buffer with some choice.
+    """
+
+    @stateless(restore_seeds=True)
+    def encode_sample(self, ...):
+        # Added `stateless` decorator to allow saving samples for slicing. Will set the seed
+        # deterministically based on the self.current_sample_index.
+        ...
+
+    def slice_batch(self, samples: List[CaptioningSample]) -> List[List[CaptioningSample]]:
```
philipp-fischer, line 37: As discussed we might want to change how slicing and batching interact. We should make sure that we can use a batch_size > 1 even when using slicing (on top).

philipp-fischer (changes requested):

src/megatron/energon/wrappers/slice_batch_dataset.py
```diff
@@ -0,0 +1,337 @@
...
+    _slicer_sample_index: SampleIndex
+    #: Sample index for the batcher function
+    _batcher_sample_index: SampleIndex
+
+    def __init__(
+        self,
+        dataset: SavableDataset[T_batch_sample],
+        buffer_size: int,
+        slicer: Callable[[List[T_batch_sample]], List[List[T_batch_sample]]],
+        batcher: Callable[[List[T_batch_sample]], T_batch],
```
voegtlel, line 89: TODO: Change this to `packer`, such that the batcher can be applied on top

src/megatron/energon/task_encoder/base.py
```diff
@@ -207,6 +256,17 @@ def _batch(
         else:
             raise ValueError("Unrecognized result type.")
 
+    def slice_batch(self, samples: List[T_sample]) -> List[List[T_sample]]:
```
voegtlel, line 259: When changing the "slice_batcher", this now operates on the "slice_buffer", thus rename this to "slice"

voegtlel: TODO: Add slice buffer size, must be set somehow. Default to `0` to disable?

src/megatron/energon/task_encoder/base.py
```diff
@@ -229,19 +292,42 @@ def build_batch(
...
-            dataset = BatchDataset(
-                dataset,
-                batch_size=batch_size,
-                batcher=self.batch,
-                drop_last=batch_drop_last,
-                worker_config=worker_config,
-            )
+            if getattr(self.slice_batch, "__func__", None) is not TaskEncoder.slice_batch:
+                assert not batch_drop_last, "batch_drop_last is not supported if slicer is set"
+                dataset = SliceBatchDataset(
```
voegtlel, line 297: TODO: new: keep batch dataset, just insert the SlicePackDataset if enabled (buffer>0 and Method overridden).

src/megatron/energon/task_encoder/base.py
```diff
@@ -100,6 +102,50 @@ def batch_list(batch: List[Any]) -> Any:
...
+        return lambda f: stateless(f, restore_seeds=restore_seeds)
+    if restore_seeds:
+        worker_seed = None
+
+        @functools.wraps(fn)
+        def seed_wrapper(self, *args, **kwargs):
+            nonlocal worker_seed
+            if worker_seed is None:
+                worker_seed = WorkerConfig.active_worker_config.worker_seed()
+            seed = hash((worker_seed, self.current_sample_index))
```
voegtlel, line 138: TODO: We need a deterministic hash function here. Fix this

src/megatron/energon/task_encoder/base.py
```diff
@@ -100,6 +102,50 @@ def batch_list(batch: List[Any]) -> Any:
...
+            set_global_seeds((seed & (seed >> 32)) & 0xFFFFFFFF)
```
voegtlel, line 139: Replace `&` by `^`

tests/test_dataset.py
```diff
@@ -876,13 +893,15 @@ def test_current_batch_index(self):
         torch.manual_seed(42)
 
         class TestTaskEncoder(TaskEncoder):
```
voegtlel, line 895: TODO: Add test: Save the random numbers once, then every run the numbers must be the same, otherwise we have a breaking change if randomness across runs changes

PR #6, opened 2024-08-15
Fix dataloader being stuck iterating samples in a certain condition

voegtlel:
The dataloader becomes stuck and does not yield samples if it has a rank which has 0 samples.

This is caused by BlendDataset not being aware of empty sub-datasets, which it will try to loop infinitely (ending in an infinite `while True` without exit).

This PR fixes that condition by checking that state. Now datasets may have zero-size and that should work just fine.


src/megatron/energon/flavors/base_webdataset.py
```diff
@@ -435,15 +436,14 @@ def _shard_workers(
...
-                (num_workers * worker_config.rank + local_worker_idx)
-                * samples_per_worker
-            )
+            int((num_workers * worker_config.rank + local_worker_idx) * samples_per_worker)
             for local_worker_idx in range(num_workers + 1)
         ]
 
         return list(
-            cls._split_shards(
+            # Filter out empty shards
```
philipp-fischer, line 444:
```suggestion
            # Filter out any empty shards for this worker
```

PR #12, opened 2024-09-11
Introduce sequence packing and sample restore

philipp-fischer:
This PR is based on https://github.com/NVIDIA/Megatron-Energon/pull/5 and includes the features introduced there.

However, the `SliceBatchDataset` has been replaced by a `PackingDataset`.
While `SliceBatchDataset` mixed packing and batching together and allowed for a batch size of only `1`, the `PackingDataset` separates the two concepts. One can now pack samples from arbitrarily sized buffers and then batch them to batches of an independent batch size.


src/megatron/energon/worker.py
```diff
@@ -218,5 +259,24 @@ def worker_log(self, data: dict) -> None:
...
+        "numpy": numpy.random.get_state(),
+        "random": random.getstate(),
+    }
+
+
+def restore_global_rng_state(state: Dict[str, Any]) -> None:
+    """Restores the global rng state for torch, numpy and random."""
+    torch.set_rng_state(state["torch"])
+    numpy.random.set_state(state["numpy"])
+    random.setstate(state["random"])
```
voegtlel, line 282: Rather use a dataclass here? It's a dict with fixed keys.

Then, also make this methods?

```suggestion
@dataclass
class RandomState:
    torch: bytes # or whatever the type is
    numpy: bytes # or whatever the type is
    random: bytes # or whatever the type is

    @staticmethod
    def set_global_seeds(seed: int):
        """Sets the global seeds for torch, numpy and random."""
        torch.manual_seed(seed)
        numpy.random.seed(seed)
        random.seed(seed)

    @staticmethod
    def save_global_rng_state() -> 'RandomState':
        """Saves the global rng state for torch, numpy and random."""
        return RandomState(
            torch=torch.get_rng_state(),
            numpy=numpy.random.get_state(),
            random=random.getstate(),
        )

    def restore_global_rng_state(self) -> None:
        """Restores the global rng state for torch, numpy and random."""
        torch.set_rng_state(self.torch)
        numpy.random.set_state(self.numpy)
        random.setstate(self.random)
```

src/megatron/energon/wrappers/packing_dataset.py
```diff
@@ -0,0 +1,364 @@
...
+    _pre_packing_sample_index: SampleIndex
+
+    #: Sample index for the final_packer
+    _final_packing_sample_index: SampleIndex
+
+    def __init__(
+        self,
+        dataset: SavableDataset[T_sample],
+        buffer_size: int,
+        pre_packer: Callable[[List[T_sample]], List[List[T_sample]]],
```
voegtlel, line 96: I find `pre_packer` and `final_packer` very unclear. Suggesting `grouper` and `merger`, correspondingly in the TE naming this `pack_group` and `pack_merge`.

src/megatron/energon/task_encoder/base.py
```diff
@@ -100,6 +109,125 @@ def batch_list(batch: List[Any]) -> Any:
...
+                    restore_global_rng_state(outer_rng_state)
+                
+                # Now yield the sample.
+                # This will give control back to the caller who may
+                # change the random state.
+                yield sample
+                
+                # Save outer random state after yielding
+                outer_rng_state = save_global_rng_state()
+
```
voegtlel, line 194: I think autoformatting was not run here?

src/megatron/energon/task_encoder/base.py
```diff
@@ -100,6 +109,125 @@ def batch_list(batch: List[Any]) -> Any:
...
+                
+
+        @functools.wraps(fn)
+        def seed_wrapper(self, *args, **kwargs):
+            nonlocal worker_seed
+            if worker_seed is None:
+                worker_seed = WorkerConfig.active_worker_config.worker_seed()
+
+            # Use a deterministic hash function to compute the seed
+            hash_digest = hashlib.sha1(
```
voegtlel, line 204: Can this also be moved to the seeding code? I believe we have it in more places, e.g. where we initialize the seeds

src/megatron/energon/task_encoder/base.py
```diff
@@ -229,19 +392,51 @@ def build_batch(
                 worker_config=worker_config,
             )
         elif batch_size is not None:
+            # Check if packing methods are provided by the user
+            pre_pack_provided = getattr(self.pre_pack, "__func__", None) is not TaskEncoder.pre_pack
```
voegtlel, line 396: duplicate, already in line 373

src/megatron/energon/task_encoder/base.py
```diff
@@ -207,19 +338,51 @@ def _batch(
...
+    def pre_pack(self, samples: List[T_sample]) -> List[List[T_sample]]:
+        """
+        For packing, selects the samples to be packed together. By default, no packing is happening.
+
+        Args:
+            samples: The samples to pre-pack. A full buffer will be passed into the function.
+
+        Returns: The pre-packed samples as a list of lists of samples.
+        """
+        raise NotImplementedError("Packing only effective when overridden.")
```
voegtlel, line 350: I suggest a default identity implementation instead of raising?
```py
return [[s] for s in samples]
```

src/megatron/energon/task_encoder/base.py
```diff
@@ -207,19 +338,51 @@ def _batch(
...
+    def final_pack(self, samples: List[T_sample]) -> T_sample:
+        """
+        Given one set of samples to pack, returns the final packed sample. By default, no packing is happening.
+
+        Args:
+            samples: The samples to pack into a single sample
+
+        Returns: The final packed sample.
+        """
+        raise NotImplementedError("Packing only effective when overridden.")
```
voegtlel, line 361: Same, default identity.
```py
assert len(samples) == 1
return samples[0]
```

src/megatron/energon/task_encoder/base.py
```diff
@@ -207,19 +338,51 @@ def _batch(
...
+        )
+
         if (
             getattr(self.batch_group_criterion, "__func__", None)
             is not TaskEncoder.batch_group_criterion
         ):
             assert batch_size is not None, "batch_size must be set if batch_group_criterion is set"
+            assert (
+                not pre_pack_provided and not final_pack_provided
+            ), "Packing not supported when grouping"
```
voegtlel, line 385: also assert that `packing_buffer_size is None`

src/megatron/energon/task_encoder/base.py
```diff
@@ -229,19 +392,51 @@ def build_batch(
...
+                    dataset,
+                    self.encode_batch,
+                    worker_config=worker_config,
+                    stateless_map_fn=getattr(self.encode_batch, "__stateless__", False),
+                )
         else:
             assert (
                 getattr(self.encode_batch, "__func__", None) is TaskEncoder.encode_batch
             ), "batch_size is not set, but encode_batch is not the default."
+            assert (
```
voegtlel, line 437: `assert packing_buffer_size is None`

voegtlel (changes requested):

src/megatron/energon/worker.py
philipp-fischer: Yeah we discussed this. I forgot about it. Will do

src/megatron/energon/task_encoder/base.py
philipp-fischer: I think we don't want to wrap the dataset in a PackingDataset if those methods are not defined by the user

voegtlel: Yes, I suggest to keep the code below which will not use it. But still define the noop

src/megatron/energon/task_encoder/base.py
philipp-fischer: Yeah good point. And I also want to allow the user to disable packing by setting `packing_buffer_size` to `None` even when the two methods are defined in the TaskEncoder. 
Tuomas I think asked how he can disable packing.

src/megatron/energon/task_encoder/base.py
philipp-fischer: We don't want this method to be called ever. If this implementation is called, it's an error, hence we must raise it.

src/megatron/energon/task_encoder/base.py
philipp-fischer: Hm, actually packing should be supported even without batching I guess.

voegtlel (approved):

PR #21, opened 2024-10-24
info command

voegtlel:
Add an info command to quickly get details of a dataset


philipp-fischer:
Good olde info command 🙏

PR #32, opened 2024-11-27
Improve distribution of samples to workers when using many datasets

philipp-fischer:
When blending many datasets together, the samples of each dataset will be distributed across all workers available globally.
Since the number of samples is typically not divisible by the number of workers, some workers will obtain more samples than others.
In the case of many datasets, this imbalance may be amplified if each dataset assigns the remainder samples to the same workers. This is leading to some workers having significantly more samples than others.

This PR mitigates this effect by shifting the assignment of samples to workers in a way that minimizes the imbalance. To that end, each dataset starts assigning samples to workers at an offset that depends on the lengths of the previous datasets, essentially filling up workers equally. 
The order in which the filling up happens, is not left-to-right, but given by a divide-and-interleave algorithm that corresponds to a generalized bit reversal permutation. This helps balance out the remainder samples across ranks in a rank-count-independent way.

BREAKING CHANGE: By improved balancing, the data ordering will not be exactly the same as before. The difference is more severe when many datasets are blended.

### DIFF
--- a/src/megatron/energon/flavors/webdataset/sharder.py
+++ b/src/megatron/energon/flavors/webdataset/sharder.py
@@ -235,60 +235,62 @@ class Sharder:
         num_workers_with_more_samples = total_samples % global_workers
 
         # We are going to compute the samples assigned to each worker on the current rank.
-        # First we compute the global worker sample assignments (rotated by the rotation offset).
-        # Then we extract the local worker sample assignments for the current rank.
+        # This is done in multiple steps.
+        # Some of these steps could be collapsed into one, but we keep them separate for clarity:
+        # 1. Compute the number of samples per global worker (rotated by rotation_offset,
+        #    typically given by previous datasets).
+        # 2. Permute the nuber of samples per global worker by a generalized bit reversal sequence
+        # 3. Given the sample counts, compute the start and end indices for each global worker
+        # 4. Extract the local worker sample assignments for the current rank.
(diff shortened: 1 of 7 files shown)


src/megatron/energon/flavors/webdataset/sharder.py
```diff
@@ -181,21 +231,78 @@ def shard_workers(
...
+
+        # 3. Compute the global worker sample start and end indices
+        global_worker_sample_start_ends = []
+        cur_offset = 0
+        for global_worker_idx in range(global_workers):
+            next_offset = cur_offset + num_samples_per_global_worker[global_worker_idx]
+            global_worker_sample_start_ends.append((cur_offset, next_offset))
+            cur_offset = next_offset
 
-        # Let's look at the workers of the current local rank.
```
voegtlel, line 186: I believe, this can be reverted now, as well as the `_split_shards` function. We already have speed issues with this part for lots of shards, so maybe rather keep it optimized, and not remove optimization here?

src/megatron/energon/flavors/webdataset/sharder.py
```diff
@@ -60,18 +61,19 @@ def _split_shard(
     def _split_shards(
```
voegtlel, line 61: Can be reverted now, that worker slices are consecutive again.

voegtlel (changes requested):

PR #33, opened 2024-11-29
Fix zero worker determinism with custom task encoder randomness

### LINKED ISSUE #23: Fix randomness when num_workers=0
When using no data loader workers at all, the `_init_worker` method is not called and hence the RNG is not seeded accordingly.
Since data loading happens in the main process, we need to save and restore all RNG states before and after calling any user code from the TaskEncoder in this case.

philipp-fischer:
Demonstrates that #23 is already resolved when using `@stateless(restore_seeds=True)`


voegtlel (approved):
Only adds tests, so good for me

PR #37, opened 2025-01-03
Refactor Docs

philipp-fischer:
This PR refactors the documentation and updates almost all sections in the docs.


docs/source/advanced/joining_datasets.md
```diff
@@ -76,6 +88,7 @@ class JoinedSample(Sample):
 ```
 
 This class should implement the `from_joined` method to combine samples from `ds1` and `ds2`.
+Note: It is important to use `derive_from` with the first argument being the first sample, as this will guarantee that the state can be saved and restored. It ensures that all the internal keys of the sample are retained.
 
 ## Example Structure
```
voegtlel, line 93: I'd suggest to move the structure up to the example scenario

PR #39, opened 2025-01-13
Fixed size formatting when reaching edge case of size > PiB

nvnbagrov:


philipp-fischer:
@voegtlel Looks good to me

PR #40, opened 2025-01-14
Change default for gc freeze to True for much higher speed

voegtlel:
Freezing the python memory before starting, improves speed for garbage collection a lot. Some tests run about 5x faster.

### DIFF
--- a/src/megatron/energon/loader.py
+++ b/src/megatron/energon/loader.py
@@ -3,12 +3,9 @@
 
 from typing import Optional, TypeVar
 
-from torch.utils.data import DataLoader
-
 from megatron.energon.flavors.base_dataset import SavableDataset
 from megatron.energon.savable_loader import BasicDataLoader, SavableDataLoader
 from megatron.energon.worker import WorkerConfig
-from megatron.energon.wrappers.gc_dataset import GcDataset
 
 T = TypeVar("T")
 
@@ -91,4 +88,4 @@ def get_loader(
                 DeprecationWarning,
             )
 
-    return BasicDataLoader(GcDataset(dataset, worker_config=dataset.worker_config))
+    return BasicDataLoader(dataset)
--- a/src/megatron/energon/wrappers/gc_dataset.py
+++ b/src/megatron/energon/wrappers/gc_dataset.py
@@ -121,4 +121,4 @@ class GcDataset(BaseSingleWrapperDataset[T_sample, T_sample], Generic[T_sample])
         return self.dataset.config()
 
     def __str__(self):
-        return f"GcDataset(every_n_iter={self.every_n_iter}, dataset={self.dataset})"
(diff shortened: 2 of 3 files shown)


src/megatron/energon/wrappers/gc_dataset.py
```diff
@@ -121,4 +121,4 @@ def config(self) -> Dict[str, Any]:
         return self.dataset.config()
 
     def __str__(self):
-        return f"GcDataset(every_n_iter={self.every_n_iter}, dataset={self.dataset})"
+        return f"GcDataset(freeze={self.freeze}, every_n_iter={self.every_n_iter}, dataset={self.dataset})"
```
philipp-fischer, line 124: Sugggest same order as actual args

src/megatron/energon/loader.py
```diff
@@ -91,4 +88,4 @@ def get_loader(
                 DeprecationWarning,
             )
 
-    return BasicDataLoader(GcDataset(dataset, worker_config=dataset.worker_config))
```
philipp-fischer, line 94: Was this a bug?

src/megatron/energon/wrappers/gc_dataset.py
```diff
@@ -75,7 +75,7 @@ def __init__(
         *,
         worker_config: WorkerConfig,
         every_n_iter: int = 1,
-        freeze: bool = False,
+        freeze: bool = True,
```
philipp-fischer, line 78: When the worker process starts, we are calling `gc_init_worker` which freezes the tensors. So why do we need to freeze again in the beginning of `__iter__`?

voegtlel: Because frozen objects are not considered by the garbage collector, which makes the calls to `gc.collect` much faster. While the objects in `gc_init_worker` are still considerend by the gc.

src/megatron/energon/loader.py
voegtlel: Yes

src/megatron/energon/wrappers/gc_dataset.py
philipp-fischer: Ok got it

philipp-fischer: We could call `gc.freeze()` at the end of `gc_init_worker`?

src/megatron/energon/wrappers/gc_dataset.py
voegtlel: Okkkay

src/megatron/energon/wrappers/gc_dataset.py
voegtlel: Well, shortly after, when the `GcDataset` kicks in 🤷 I think this does not make much difference, only that it would not be configurable as easily in the `gc_init_worker`?

PR #41, opened 2025-01-14
Accurate Metadataset Epoch

### LINKED ISSUE #14: Accurate Metadataset Epoch
For LLMs, it is a common methodology, to show every sample exactly once (or n times). The metadataset currently only implements blending/mixing with probability. As a workaround, the sampling probability could be set to the size of the respective dataset, but that doesn't ensure that every sample is seen exactly once (considering shuffle buffer, and repeating) and it doesn't stop the dataset iteration.

A suggestion for an additional interface:

```py
# epochized=True makes the dataset not repeat, but uses the underlying metadataset definition to stop iterating. Needs handling for non-complete batches
def get_train_dataset(..., repeat: bool = True):
```

And in the dataset:
```yaml
__module__: megatron.energon
__class__: MetadatasetV2
splits:
  train:
    # If blend_epochized is used, do not weight the dataset, but use repetitions for, i.e. the weight is effectively repetitions multiplied with the dataset size
    blend_epochized:
      - repetitions: 1
        path: ...
      - ...
```

If `weight_samples=True` (also for nested datasets), and the dataset weights are all integers, then `epochized=True` may be set for `get_train_dataset`, implicating the following constraint: `get_train_dataset` stops iterating after every sample in the datasets has been seen respective `weight` times.

Implementation thoughts:
* Set `repeat=False` for the `WebdatasetSampleLoaderDataset`. Maybe this parameter can be removed entirely. Makes the inner dataset stop after one epoch.
* Implement new `BlendDataset` wrapper, which ends after one epoch and does not loop inner datasets

voegtlel:
Fixes #14

Contains a (checkpoint) breaking change, which makes existing checkpoints incompatible, because the `RepeatDataset` is now inserted into the tree.


src/megatron/energon/wrappers/blend_dataset.py
```diff
@@ -65,16 +65,13 @@ def __init__(
         self._worker_rng = WorkerRng(self.worker_config)
 
     def __len__(self) -> int:
+        return sum(len(dataset) for dataset, weight in self.dataset_weights)
```
voegtlel, line 69: I believe this shall not have an effect

src/megatron/energon/wrappers/blend_dataset.py
```diff
@@ -85,14 +77,26 @@ def __iter__(self) -> Iterator[T_sample]:
...
-            sample = next(dataset_iters[ds_idx])
-            yield add_sample_restore_key(sample, ds_idx, src=self)
+            if dataset_iters[ds_idx] is None:
+                if all(dataset_iter is None for dataset_iter in dataset_iters):
+                    break
+                continue
+            try:
+                sample = next(dataset_iters[ds_idx])
+            except StopIteration:
+                dataset_iters[ds_idx] = None
```
voegtlel, line 94: This can now stop iterating when the inner datasets are exhausted

src/megatron/energon/wrappers/epochize_dataset.py
```diff
@@ -56,11 +56,6 @@ def __init__(
         self._offset = [0] * max(self.worker_config.num_workers, 1)
         self._active_iter = None
 
-    def _infinite(self) -> Generator[T_sample, None, None]:
```
voegtlel, line 59: Make this end when the inner dataset ends

src/megatron/energon/wrappers/gc_dataset.py
```diff
@@ -75,7 +75,7 @@ def __init__(
         *,
         worker_config: WorkerConfig,
         every_n_iter: int = 1,
-        freeze: bool = False,
+        freeze: bool = True,
```
voegtlel, line 78: Should first merge #40

src/megatron/energon/wrappers/repeat_dataset.py
```diff
@@ -86,6 +91,8 @@ def __iter__(self) -> Iterator[T_sample]:
                             "repeats": self.repeats,
                         }
                     )
+            if self.restart:
+                self._offset[worker_idx] = 0
```
voegtlel, line 95: Must reset to be able to continue iterating

tests/test_metadataset_v2.py
```diff
@@ -473,6 +477,144 @@ def test_joined_metadataset_joiner(self):
...
+        worker_config = WorkerConfig(
+            rank=0,
+            world_size=1,
+            num_workers=0,
+            seed_offset=42,
+        )
+
+        # Create a joined dataset configuration
+        fixed_epochs_mds_path = self.dataset_path / "metadataset_fixed_epochs.yaml"
+        with open(fixed_epochs_mds_path, "w") as f:
```
voegtlel, line 491: Test 2x ds1, 3x ds2 and verify that stop iteration is working correctly, with save/restore

PR #42, opened 2025-01-14
License header checker

voegtlel:
Checks that all files have correct licenses and can fix the headers

### DIFF
--- a/scripts/license_headers.py
+++ b/scripts/license_headers.py
@@ -1,0 +1,113 @@
+# Copyright (c) 2025, NVIDIA CORPORATION.
+# SPDX-License-Identifier: BSD-3-Clause
+from dataclasses import dataclass
+from pathlib import Path
+import click
+
+
+@dataclass
+class HeaderUpdater:
+    file_ext: str
+    line_comment: str | None = None
+    comment_start: str | None = None
+    comment_end: str | None = None
+
+    UPDATE_IDENTIFIER = "Copyright"
+
+    HEADER_LINES: tuple[str, ...] = (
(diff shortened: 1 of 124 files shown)


.github/workflows/black.yml
```diff
@@ -0,0 +1,33 @@
...
+      with:
+        python-version: '3.10'
+
+    - name: Install dependencies
+      run: |
+        python -m pip install --upgrade pip
+        pip install black
+
+    - name: Run Black
+      run: black --check .
```
philipp-fischer, line 33: Did you verify it uses our pyproject.toml? I guess it should

.github/workflows/black.yml
```diff
@@ -0,0 +1,33 @@
+# Copyright (c) 2025, NVIDIA CORPORATION.
```
philipp-fischer, line 1: We don't need copyrght on workflows... Source code files should be sufficient

scripts/license_headers.py
```diff
@@ -0,0 +1,122 @@
...
+headers = (
+    HeaderUpdater(
+        file_ext=".py",
+        line_comment="# ",
+    ),
+    HeaderUpdater(
+        file_ext=".sh",
+        line_comment="# ",
+    ),
+    HeaderUpdater(
```
philipp-fischer, line 76: Like stated above, I wouldn't do this on yamls

philipp-fischer (changes requested):

scripts/license_headers.py
voegtlel: Okok

philipp-fischer:
lgtm

PR #43, opened 2025-01-15
Cleanup SavableLoader's save and restore state methods

voegtlel:
* Make `save_state`/`restore_state` deprecated and rename to `save_state_global`/`restore_state_global` to make usage more clear
* Update docs
* Tiny fixing/breaking change: Removed `save_state(dst_rank=None)` option, because there is no useful scenario for that and it's confusing. It will break code which was using it wrong that way.


src/megatron/energon/savable_loader.py
```diff
@@ -944,89 +944,110 @@ def restore_state_rank(self, state: Optional[SavableDataLoaderState]) -> None:
...
-    def save_state(
-        self, dst_rank: Optional[int] = None
-    ) -> Union[SavableDataLoaderState, List[SavableDataLoaderState], None, List[None]]:
+    def save_state(self, dst_rank: int) -> List[Optional[SavableDataLoaderState]]:
+        """Deprecated. Use `save_state_global` (or `save_state_rank`) instead."""
+        warnings.warn(
+            "`save_state` is deprecated and was renamed to `save_state_global` and will be removed "
+            "in a future update. If you actually do not want to gather the states to a rank, use "
+            "`save_state_rank` instead.",
+            DeprecationWarning,
```
philipp-fischer, line 966: Docs say this is filtered out by default. Did you check? See https://docs.python.org/3/library/exceptions.html#DeprecationWarning

src/megatron/energon/savable_loader.py
```diff
@@ -944,89 +956,113 @@ def restore_state_rank(self, state: Optional[SavableDataLoaderState]) -> None:
...
-                )
-                state = local_object[0]
-        self.restore_state_rank(state)
+        if src_rank is None or self.worker_config.world_size == 1:
+            # All ranks have the state
+            # Select the state of the current rank
+            rank_state = state[self.worker_config.rank]
+        else:
+            # Only the src_rank has the state
+            if self.worker_config.rank != src_rank:
```
philipp-fischer, line 1051: Bug? Below we have `assert state is None` for non-source ranks but above for all ranks we have `assert isinstance(state, list)`

philipp-fischer: I think I fixed it, but let's take a look together

src/megatron/energon/savable_loader.py
philipp-fischer: I tried and it doesn't show up at all when raised in an imported lib like energon

src/megatron/energon/savable_loader.py
voegtlel: lgtm

PR #49, opened 2025-01-22
Reduce checkpoint size by not saving the pending shards

voegtlel:
Just save, how the pending shards were generated (i.e. the random state for shuffling and the number of shards that have already been processed)
(contains CHECKPOINT BREAKING CHANGE)

### DIFF
--- a/src/megatron/energon/flavors/webdataset/sample_loader.py
+++ b/src/megatron/energon/flavors/webdataset/sample_loader.py
@@ -43,9 +43,13 @@ class SampleLoaderState(State):
 
     #: Rng state
     rng: WorkerRngState
-    #: Pending shards are the shards which have not yet been opened, but should be processed
-    # in the current "epoch"
-    pending_shards: Optional[List[List[ShardInfo]]]
+    #: The seed that was used to generate the current pending shards
+    pending_shards_seed: WorkerRngState
+    #: The number of shards that have already been opened / processed and thus been removed from the
+    # pending shards. None if starting a new epoch.
+    # Pending shards are the shards which have not yet been opened, but should be processed in the
+    # current epoch.
+    pending_shards_offset: Optional[int]
     #: The active shards are the currently opened shards. May contain `None`, if there are fewer
     # shards available (i.e. pending_shards empty) than parallel shard iterators requested.
     active_shards: Optional[List[Optional[List[ShardState]]]]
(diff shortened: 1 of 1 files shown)


src/megatron/energon/flavors/webdataset/sample_loader.py
```diff
@@ -701,16 +765,11 @@ def restore_state(self, state: Optional[SampleLoaderMergedState]) -> None:
             assert isinstance(state, SampleLoaderMergedState)
             self._worker_rng.restore_state(state.rng)
             # Restore state
-            assert len(state.pending_shards) == len(self.shards)
+            assert len(state.pending_shards_offset) == len(self.shards)
             assert len(state.active_shards) == len(self.shards)
-            self._pending_shards = [
```
philipp-fischer, line 706: I guess we will have a CHECKPOINT BREAKING CHANGE, although in principle we could also restore and old checkpoint here, right?

src/megatron/energon/flavors/webdataset/sample_loader.py
```diff
@@ -259,52 +274,84 @@ def _read_multicolumn_shard(
...
+                res_list = [
+                    shards[rng.randbelow(len(shards))] for _ in range(self.parallel_shard_iters)
+                ]
+            elif self.shuffle_over_epochs >= 1:
+                # Shuffle without replacement (potentially over multiple epochs)
+                res_list = rng.shuffle(shards * self.shuffle_over_epochs)
+            else:
+                raise ValueError(f"Invalid shuffle_over_epochs: {self.shuffle_over_epochs}")
+        # Reverse, such that pop returns the first element (in O(1) time)
+        res_list.reverse()
```
philipp-fischer, line 313: Aren't you reversing twice now? One more time after `_shards_once` is called.

philipp-fischer: Ok I'll take it back, you removed that in the outer scope

src/megatron/energon/flavors/webdataset/sample_loader.py
```diff
@@ -391,8 +438,10 @@ def _shards_iter(
             # List of shard iterators, always of length `parallel_shard_iters`. May contain `None`.
             active_shards = []
             # Fill up the shard iterators
-            while len(shards_order) > 0 and len(shard_iters) < self.parallel_shard_iters:
-                subshards = shards_order.pop()
+            while len(pending_shards) > 0 and len(shard_iters) < self.parallel_shard_iters:
+                subshards = pending_shards.pop()
+                assert self._pending_shards_offset[worker_idx] is not None
+                self._pending_shards_offset[worker_idx] += 1
```
philipp-fischer, line 444: Is the this case covered in a unit test? I mean the case where we save/restore after more than 1 epoch and get another shard ordering

philipp-fischer (commented):
lgtm

src/megatron/energon/flavors/webdataset/sample_loader.py
voegtlel: I think so, yes. I remember having written such tests. `test_metadataset.test_save_restore_state_train_workers`

src/megatron/energon/flavors/webdataset/sample_loader.py
voegtlel: Yes. But a CHECKPOINT BREAKING CHANGE is just a minor version increase :)

src/megatron/energon/flavors/webdataset/sample_loader.py
voegtlel: Wasn't easy to find 🙈 Maybe we should cleanup tests and document them better

voegtlel:
Already solved in #60

PR #57, opened 2025-02-06
Refactored `prepare`

nvnbagrov:
Simplified code and logic
* No dummy class
* No conditional logic

### DIFF
--- a/src/megatron/energon/tools/prepare.py
+++ b/src/megatron/energon/tools/prepare.py
@@ -245,22 +245,21 @@ def command(
         # Get a list of all classes in megatron.energon that are subclasses of WebdatasetBase
         import megatron.energon as data_import
 
-        name_to_class = {
-            name: cls
+        display_name_and_class = [
+            (name, cls)
             for name, cls in inspect.getmembers(data_import)
             if isinstance(cls, type) and issubclass(cls, Sample)
-        }
-        name_to_class["Crude sample (plain dict for cooking)"] = CrudeWebdataset
-        name_to_class = OrderedDict(name_to_class)
+        ]
+        display_name_and_class.append(("Crude sample (plain dict for cooking)", CrudeWebdataset))
 
         # Print all classes and ask user to pick one
         click.echo("The following sample types are available:")
-        for i, (name, cls) in enumerate(name_to_class.items()):
+        for i, (name, cls) in enumerate(display_name_and_class):
             click.echo(f"{i}. {name}")
         while True:
             choice = click.prompt("Please enter a number to choose a class", type=int)
             try:
-                cls = name_to_class[list(name_to_class.keys())[choice]]
+                _, cls = display_name_and_class[choice]
                 break
             except IndexError:
                 click.echo("Invalid choice. Please try again.")
--- a/src/megatron/energon/tools/prepare.py
+++ b/src/megatron/energon/tools/prepare.py
@@ -6,6 +6,7 @@ import inspect
 import json
 import re
 import typing
+from collections import OrderedDict
 from types import FunctionType
 from typing import Any, List, Optional, Type
 
@@ -16,10 +17,6 @@ from megatron.energon.epathlib import EPath
 from megatron.energon.flavors import BaseWebdatasetFactory, CrudeWebdataset
 from megatron.energon.flavors.base_dataset import Sample
 from megatron.energon.flavors.webdataset import MAIN_FOLDER_NAME
-
-
-class CrudeSampleDummy:
-    pass
 
 
 def type_str(tp: Type) -> str:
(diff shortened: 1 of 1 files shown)


src/megatron/energon/tools/prepare.py
```diff
@@ -18,10 +19,6 @@
 from megatron.energon.flavors.webdataset import MAIN_FOLDER_NAME
 
 
-class CrudeSampleDummy:
```
voegtlel, line 21: Agreed, that this can be removed

src/megatron/energon/tools/prepare.py
```diff
@@ -248,56 +245,46 @@ def progress_fn(els, length=None):
...
 
-        all_classes = []
-        for name, cls in inspect.getmembers(data_import):
-            if isinstance(cls, type) and issubclass(cls, Sample):
-                all_classes.append(cls)
-
-        all_classes.append(
-            ("Crude sample (plain dict for cooking)", CrudeSampleDummy)
-        )  # Tuple is (Printed name, resulting class)
+        name_to_class = {
```
voegtlel, line 248: Of course, this can be made a comprehension, but I don not see the advantage of the dictionary?

src/megatron/energon/tools/prepare.py
```diff
@@ -248,56 +245,46 @@ def progress_fn(els, length=None):
...
+            name: cls
+            for name, cls in inspect.getmembers(data_import)
+            if isinstance(cls, type) and issubclass(cls, Sample)
+        }
+        name_to_class["Crude sample (plain dict for cooking)"] = CrudeWebdataset
+        name_to_class = OrderedDict(name_to_class)
```
voegtlel, line 254: Since python 3.7, dicts are already ordered, so this is not needed.

src/megatron/energon/tools/prepare.py
```diff
@@ -248,56 +245,46 @@ def progress_fn(els, length=None):
...
-                click.echo(f"{i}. {cls[0]}")
-            else:
-                click.echo(f"{i}. {cls.__name__}")
+        for i, (name, cls) in enumerate(name_to_class.items()):
+            click.echo(f"{i}. {name}")
         while True:
             choice = click.prompt("Please enter a number to choose a class", type=int)
             try:
-                cls = all_classes[choice]
+                cls = name_to_class[list(name_to_class.keys())[choice]]
```
voegtlel, line 263: This looks much more complicated than before? Why not keep the list of tuples?

src/megatron/energon/tools/prepare.py
```diff
@@ -248,56 +245,46 @@ def progress_fn(els, length=None):
...
                 break
             except IndexError:
                 click.echo("Invalid choice. Please try again.")
                 continue
 
-        if isinstance(cls, tuple):
-            cls = cls[1]
+        sample_type = {
```
voegtlel, line 269: Why revert this? I think the current version where this is moved inside is better, because we need the two cases anyways. I think this is now rather confusing, than having the two cases clearly separated

voegtlel (changes requested):
Hi @nvnbagrov, thanks for this PR! I think we can keep the removal of `CrudeSampleDummy`, but I'd revert the changes of using a dict and moving the definition out of the `if cls == CrudeWebdataset`.

src/megatron/energon/tools/prepare.py
nvnbagrov: The cases are clearly separated, `dataset_definition` is defined twice (both `if` and `else`). LMK if you wish to drop `sample_type` and revert back to the old definition.

src/megatron/energon/tools/prepare.py
nvnbagrov: Thanks, simplified

[using list of tuples and not ordered dict](https://github.com/NVIDIA/Megatron-Energon/pull/57/commits/a85ccd6857416924319503e5944ca4ad4681138e)

src/megatron/energon/tools/prepare.py
nvnbagrov: Thanks, irrelevant

[using list of tuples and not ordered dict](https://github.com/NVIDIA/Megatron-Energon/pull/57/commits/a85ccd6857416924319503e5944ca4ad4681138e)

src/megatron/energon/tools/prepare.py
nvnbagrov: Thanks, 
[using list of tuples and not ordered dict](https://github.com/NVIDIA/Megatron-Energon/pull/57/commits/a85ccd6857416924319503e5944ca4ad4681138e)

src/megatron/energon/tools/prepare.py
```diff
@@ -248,56 +245,45 @@ def progress_fn(els, length=None):
...
-
-        all_classes.append(
-            ("Crude sample (plain dict for cooking)", CrudeSampleDummy)
-        )  # Tuple is (Printed name, resulting class)
+        display_name_and_class = [
+            (name, cls)
+            for name, cls in inspect.getmembers(data_import)
+            if isinstance(cls, type) and issubclass(cls, Sample)
+        ]
+        display_name_and_class.append(("Crude sample (plain dict for cooking)", CrudeWebdataset))
```
nvnbagrov, line 253: LMK if I should keep the original `for`-loop logic.

voegtlel (changes requested):

src/megatron/energon/tools/prepare.py
nvnbagrov: Done, [moved definition inside conditional logic](https://github.com/NVIDIA/Megatron-Energon/pull/57/commits/0266f3751acb009b9ed71c351d233313f6ba557a)

voegtlel (approved):
Thanks for the changes! Looks good to me now :)

PR #64, opened 2025-02-10
fixed bug with encode_batch not being called for grouping dataset

nvnbagrov:


voegtlel (commented):
lgtm, how could that happen? 😅

nvnbagrov:
When you implement both methods, simple as that 🤷‍♂️

PR #65, opened 2025-02-11
Fix: Restore packing ability to fill the buffer on-the-fly

voegtlel:


src/megatron/energon/wrappers/packing_dataset.py
```diff
@@ -241,33 +241,30 @@ def next_final_pack() -> Generator[T_batch_sample, None, None]:
...
-
                 next_pre_pack()
+                if len(pre_packing_lengths) == 0:
+                    # Retry packing, nothing was returned.
+                    pre_pack_round += 1
+                    continue
 
-                pre_pack_round += 1
-                if pre_pack_round > 10:
-                    raise RuntimeError("Pre packer did not yield any packs after 10 rounds.")
```
philipp-fischer, line 265: We need to bring back the error message when the loop exits due to this reason

philipp-fischer (changes requested):

PR #67, opened 2025-02-12
Toolchain upgrade: Drop black, isort and introduce ruff, uv and just

philipp-fischer:
Go for more modern tools.

Things not done yet:

- Adapt doc build github workflow to use just
- Adapt release workflow to use just
- Handle code parts where ruff and black disagree (ruff currently fails)
- Update README and docs to explain tools and just-commands


pyproject.toml
```diff
@@ -12,10 +12,9 @@ authors = [
 description = "Megatron's multi-modal data loader"
 readme = "README.md"
 license = "BSD-3-Clause"
-requires-python = ">=3.8"
+requires-python = ">=3.10"
```
voegtlel, line 15: Huh, we're at 3.10 now? I thought that was still in discussion? 3.9 for sure, but I don't remember finalizing on 3.10

voegtlel:
Before we apply all the things `ruff` detects, let's merge the larger MRs, otherwise it'll be a nightmare :D

voegtlel (approved):

PR #69, opened 2025-02-13
Fix restore key if there's an empty worker

philipp-fischer:
Similar to #68, but on the `develop` branch.

### DIFF
--- a/src/megatron/energon/wrappers/blend_dataset.py
+++ b/src/megatron/energon/wrappers/blend_dataset.py
@@ -94,7 +94,7 @@ class BlendDataset(BaseWrapperDataset[T_sample], Generic[T_sample]):
                 weights.append(0)
 
         weights = torch.tensor(weights, dtype=torch.float32)
-        if probs.sum() == 0:
+        if weights.sum() == 0:
             raise RuntimeError(
                 "There is a worker with no samples in any of the blended datasets. "
                 "This can happen if you have a lot of workers and your dataset is too small. "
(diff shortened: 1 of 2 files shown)


src/megatron/energon/wrappers/blend_dataset.py
```diff
@@ -77,18 +77,31 @@ def __len__(self) -> int:
...
+            else:
+                dataset_iters.append(None)
+                weights.append(0)
+
+        probs = torch.tensor(weights, dtype=torch.float32)
+        if probs.sum() == 0:
+            raise RuntimeError(
+                "There is a worker with no samples in any of the blended datasets. "
+                "This can happen if you have a lot of workers and many small datasets. "
+                "Currently this case is not supported."
```
voegtlel, line 101:
```suggestion
                "There is a worker with no samples in any of the blended datasets. "
                "This can happen if you have a lot of workers and your dataset is too small. "
                "Currently this case is not supported."
```
With rotation, the number of datasets does not play a role any more.

src/megatron/energon/wrappers/blend_dataset.py
```diff
@@ -77,18 +77,31 @@ def __len__(self) -> int:
...
+            assert weight > 0, "All blending weights must be > 0"
+
+            if dataset.worker_has_samples():
+                dataset_iters.append(iter(dataset))
+                weights.append(weight)
+            else:
+                dataset_iters.append(None)
+                weights.append(0)
+
+        probs = torch.tensor(weights, dtype=torch.float32)
```
voegtlel, line 96: This is now `weights` again, right? `probs` would need to be normalized.

Issue #71, opened 2025-02-17
Bug in save_state_global if non-data-parallelism is active


philipp-fischer:
This line of code assumes that the rank from the worker config is unique: https://github.com/NVIDIA/Megatron-Energon/blob/develop/src/megatron/energon/savable_loader.py#L997

However, in case of e.g. tensor parallelism there can be multiple data loaders with the same rank and all of them will try to save the state globally.
To solve this, we could allow passing the current global unique rank to `save_state_global`.

philipp-fischer:
Resolved in #72

PR #72, opened 2025-02-17
Feature/parallelism

philipp-fischer:
Fix global state save/restore and improve docs around save/restore in distributed setting. Addresses https://github.com/NVIDIA/Megatron-Energon/issues/71

Also remove the src_rank argument from `restore_state_global`, since that would not work with tensor parallelism.

### DIFF
--- a/docs/source/advanced/parallelism.md
+++ b/docs/source/advanced/parallelism.md
@@ -17,7 +17,7 @@ These parallelisms have different consequences for the dataloader:
 
 - **Data Parallelism** (DP): The dataloader needs to ensure that each processor gets a different subset of the data. This is supported by Energon. The data parallel groups should be specified in the worker config.
 
-- **Pipeline Parallelism** (PP): Data is typically only loaded on the first Pipeline Parallel rank, and propagates through the other ranks within the pipeline parallel group. This means, you only instantiate an enegon dataset and loader on the first ranks of those groups.
+- **Pipeline Parallelism** (PP): Data is typically only loaded on the first Pipeline Parallel rank, and propagates through the other ranks within the pipeline parallel group. This means, you only instantiate an Energon dataset and loader on the first ranks of those groups.
 
 - **Tensor Parallelism** (TP): The dataloader will load the same input data on multiple devices. Typically, this can be ensured by either instantiating the dataloader exactly the same on the same data parallel ranks in different data parallel groups, or e.g. by loading the data only once and distributing it using torch distributed.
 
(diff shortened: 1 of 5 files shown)


docs/source/advanced/parallelism.md
```diff
@@ -0,0 +1,140 @@
...
+
+3. **Pipeline Parallelism** (PP): This technique involves breaking down the model into different stages and processing different mini-batches of data through these stages in a pipeline fashion. It helps in improving the utilization of resources and reducing idle time.
+
+4. **Tensor Parallelism** (TP): This method splits individual tensors (weights and activations) across multiple devices. It is particularly effective for very large models where even a single layer cannot fit into the memory of one device.
+
+These parallelisms have different consequences for the dataloader:
+
+- **Data Parallelism** (DP): The dataloader needs to ensure that each processor gets a different subset of the data. This is supported by Energon. The data parallel groups should be specified in the worker config.
+
+- **Pipeline Parallelism** (PP): Data is typically only loaded on the first Pipeline Parallel rank, and propagates through the other ranks within the pipeline parallel group. This means, you only instantiate an enegon dataset and loader on the first ranks of those groups.
```
voegtlel, line 20:
```suggestion
- **Pipeline Parallelism** (PP): Data is typically only loaded on the first Pipeline Parallel rank, and propagates through the other ranks within the pipeline parallel group. This means, you only instantiate an Energon dataset and loader on the first ranks of those groups.
```

docs/source/advanced/parallelism.md
```diff
@@ -0,0 +1,140 @@
...
+    # load the checkpoint from disk on every dp rank:
+    # state = torch.load("dataloader.pt")
+    # loader.restore_state_global(state)
+
+    # Or load only once from disk for each dp group:
+    # if data_parallel_rank == 0:
+    #     state = torch.load("dataloader.pt")
+    # else:
+    #     state = None
+    # loader.restore_state_global(state, src_rank=0)
```
voegtlel, line 138: here, the argument remains as `src_rank`? Shouldn't that be `global_src_rank`?

src/megatron/energon/savable_loader.py
```diff
@@ -1012,31 +1029,31 @@ def save_state_global(
...
 
-        return self.restore_state_global(state, src_rank)
+        return self.restore_state_global(state)
 
     def restore_state_global(
         self,
         state: Optional[Sequence[Optional[SavableDataLoaderState]]],
-        src_rank: Optional[int],
+        *,
+        src_rank: Optional[int] = None,
```
voegtlel, line 1041: So, this is the local rank? Should we also rename it to `src_dp_rank`? Now, this seems inconsistent with the save function?

voegtlel (changes requested):

voegtlel (approved):

PR #80, opened 2025-02-27
Refactor save/restore concept: Worker dimension as outer dimension

philipp-fischer:
This PR changes the way how the internal dataset wrappers save and restore their states.
Instead of every dataset keeping the states for all workers, they now only keep the state of the current worker.
Also, the checkpoint format changes. The saved state will contain a list of dataset states, one for each worker.
Before, the worker dimension was an inner dimension of the state of each dataset.

This will allow us to re-distribute the states easily, when resuming with a different node count but same number of global workers.

### DIFF
--- a/src/megatron/energon/flavors/base_dataset.py
+++ b/src/megatron/energon/flavors/base_dataset.py
@@ -261,6 +261,7 @@ class SavableDataset(IterableDataset[T_sample], Savable, Generic[T_sample], ABC)
         """
 
         state = FlexState()
+        state["__class__"] = type(self).__name__
         for key in self._savable_fields:
             attr = getattr(self, key)
             if isinstance(attr, Savable):
@@ -290,6 +291,9 @@ class SavableDataset(IterableDataset[T_sample], Savable, Generic[T_sample], ABC)
         Args:
             state: The state of the dataset as savable object. If None, restore initial state.
         """
+        assert state["__class__"] == type(self).__name__, (
+            f"Class name mismatch: {state['__class__']} != {type(self).__name__}"
+        )
 
         for key in self._savable_fields:
             assert key in state, f"Key {key} not in state {state}"
--- a/src/megatron/energon/flavors/base_dataset.py
+++ b/src/megatron/energon/flavors/base_dataset.py
@@ -29,6 +29,7 @@ from typing_extensions import Self
 from megatron.energon.dataclass_slots import dataclass_slots
 from megatron.energon.epathlib import EPath
 from megatron.energon.savable import Savable
+from megatron.energon.state import FlexState
 from megatron.energon.worker import WorkerConfig
 
 T_sample = TypeVar("T_sample", covariant=True)
@@ -225,10 +226,6 @@ class State(ABC, ExtendableDataclassMixin):
     """
 
 
-class FlexState(dict):
-    pass
-
-
 class SavableDataset(IterableDataset[T_sample], Savable, Generic[T_sample], ABC):
     """A dataset that can be saved and restored (i.e. the random state, internal buffers, etc.).
     I.e. it can be resumed from a checkpoint.
@@ -307,6 +304,10 @@ class SavableDataset(IterableDataset[T_sample], Savable, Generic[T_sample], ABC)
     def reset_state_own(self) -> None:
         """Resets the state of the dataset to the initial state. Can only be called in a worker process."""
         ...
+
+    def reset_state_deep(self) -> None:
+        """Resets the state of the dataset to the initial state. Can only be called in a worker process."""
+        self.reset_state_own()
 
     @abstractmethod
     def worker_has_samples(self) -> bool:
--- a/src/megatron/energon/flavors/base_dataset.py
+++ b/src/megatron/energon/flavors/base_dataset.py
@@ -267,7 +267,8 @@ class SavableDataset(IterableDataset[T_sample], Savable, Generic[T_sample], ABC)
                 state[key] = attr.save_state()
             else:
                 # Check if this field is a simple python type or a user class
-                if attr.__module__ != "builtins":
+
+                if attr is not None and getattr(attr, "__module__", "builtins") != "builtins":
                     import warnings
 
                     warnings.warn(
--- a/src/megatron/energon/flavors/base_dataset.py
+++ b/src/megatron/energon/flavors/base_dataset.py
@@ -165,9 +165,9 @@ class Sample(ABC, PinMemoryMixin, ExtendableDataclassMixin):
         Returns:
             The joined constructed sample.
         """
-        assert (
-            len(kwargs) == 0
-        ), "Please specify joined datasets as list for the default joiner. Keyword arguments are confusing, because keys are ignored."
+        assert len(kwargs) == 0, (
+            "Please specify joined datasets as list for the default joiner. Keyword arguments are confusing, because keys are ignored."
+        )
         excluded_fields = set(field.name for field in dataclasses.fields(Sample))
         init_args = {}
         if len(args) > 0:
@@ -223,46 +223,6 @@ class State(ABC, ExtendableDataclassMixin):
     """
 
 
-@dataclass_slots
-class MergedState(ABC, ExtendableDataclassMixin):
-    """An abstract base class for the merged state of a dataset. See :class:`SavableDataset`.
-    The merged state is created in the :meth:`megatron.energon.SavableDataset.merge_states` method, and
-    represents the merged state of all worker processes (only workers, not ranks). It is required
-    to restore the state of the dataset in the :meth:`megatron.energon.SavableDataset.restore_state`
-    method for all workers before the workers are started.
-    Each dataset should derive a specific subclass as a `@dataclass` and add the fields as needed
-    for training.
-
-    To extend subclasses, use the .extend method. Example:
-    Example (see :meth:`megatron.energon.State` to complete the example)
-
-    .. code-block:: python
-
-        @dataclass
-        class MyMergedState(MergedState):
-            a: List[int]
-
-        @dataclass
-        class MyExtendedMergedState(MyMergedState):
-            # Add a new field `b` to the state
-            b: List[int]
-
-        class MyMergedStateSaver:
-            def merge_state(self, states: List[MyState]) -> MyMergedState:
-                return MyMergedState(a=[s.a for s in states])
-
-        class MyExtendedMergedStateSaver(MyMergedStateSaver):
-            def merge_state(self, states: List[MyExtendedState]) -> MyExtendedMergedState:
-                # Fetch state from super class, which is already a complete instance (cannot add
-                # new fields to it, type is fixed).
-                state: MyMergedState = super().merge_state(states)
-
-                # Now extend the state of the super class (of type `MyMergedState`) with the
-                # new field required to define `MyExtendedMergedState`.
-                return MyExtendedMergedState.extend(state, b=[s.b for s in states])
-    """
-
-
 class SavableDataset(IterableDataset[T_sample], Generic[T_sample], ABC):
     """A dataset that can be saved and restored (i.e. the random state, internal buffers, etc.).
     I.e. it can be resumed from a checkpoint.
@@ -295,19 +255,6 @@ class SavableDataset(IterableDataset[T_sample], Generic[T_sample], ABC):
 
         Returns:
             The state of the dataset as savable object (i.e. python basic types).
-        """
-        ...
-
-    @abstractmethod
-    def merge_states(self, states: List[Optional[State]]) -> MergedState:
-        """
-        Merges the states of all workers into one state. Restore state expects a merged state.
-
-        Args:
-            states: The states of the workers to merge.
-
-        Returns:
-            The merged state.
         """
         ...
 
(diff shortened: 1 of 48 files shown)


src/megatron/energon/flavors/webdataset/joined_webdataset.py
```diff
@@ -186,7 +186,7 @@ def build(self, worker_rotation_offset: int = 0) -> SavableDataset[T_sample]:
 
         dataset = WebdatasetSampleLoaderDataset(
             join_readers=itar_readers,
-            workers_sample_slice_offsets=workers_sample_slice_offsets,
+            sample_slice_offsets=workers_sample_slice_offsets,
```
voegtlel, line 189: Why rename without `workers`?

src/megatron/energon/flavors/webdataset/sample_loader.py
```diff
@@ -35,60 +24,23 @@ class RawSampleData:
...
 
 class WebdatasetSampleLoaderDataset(SavableDataset[RawSampleData]):
     """Internal class for loading samples from webdataset slices"""
 
     #: The readers for each joined dataset
     join_readers: Sequence[ITarReader]
+
     #: The offsets of the slice slices to iterate over
-    worker_slice_offsets: Sequence[Sequence[int]]
+    slice_offsets: Sequence[int]
```
voegtlel, line 43: Naming is now inconsistent? I see usage of both, `workers_slice_offsets` and `slice_offsets` now. But I guess `slice_offsets` should actually be local to iter?

src/megatron/energon/savable_loader.py
```diff
@@ -485,37 +442,25 @@ def _save_state(self) -> SavableDatasetState:
...
-            sample_index=[0 if s is None else s.sample_index for s in states],
-        )
-
-    def _restore_state(self, state: SavableDatasetMergedState) -> None:
+    def _restore_state(self, state: SavableDatasetState) -> None:
         """Restores the internal worker state"""
         assert torch.utils.data.get_worker_info() is not None, "Can only restore in worker process"
-        if state.torch_rng[self._worker_id] is None:
+        if state.torch_rng is None:
             torch.manual_seed(torch.initial_seed())
```
voegtlel, line 449: Just seeing this. Is this correct? Shouldn't that rely on the configured seed nowadays? 😅

src/megatron/energon/savable_loader.py
```diff
@@ -546,21 +491,9 @@ def get_checkpoint(self, last_sample_indexes: List[int]) -> SavableDatasetCheckp
...
-        return SavableDatasetMergedCheckpoint(
-            dataset_state=self.dataset.merge_states([c.dataset_state for c in checkpoints]),
-            state=self._merge_states([c.state for c in checkpoints]),
-            offset=[c.offset for c in checkpoints],
-        )
-
     def restore_checkpoint(
         self,
-        state: Optional[SavableDatasetMergedCheckpoint],
+        state_list: Optional[List[SavableDatasetCheckpoint]],
```
voegtlel, line 496: Docs are not adapted. Also, why `state` to `state_list`? I think `list` does not help to understand that better (and typing already makes clear, it's a list :D ). Rename e.g. to `workers_state`.

src/megatron/energon/wrappers/base.py
```diff
@@ -25,132 +20,112 @@
...
 
+    def reset_state_deep(self) -> None:
+        """Resets the state of the inner datasets and then the own state."""
 
-class SampleIndex:
+        for ds in self.datasets:
+            if isinstance(ds, BaseWrapperDataset):
+                ds.reset_state_deep()
+            else:
+                ds.reset_state_own()
```
voegtlel, line 98: Why `own` if deep is requested?

src/megatron/energon/wrappers/base.py
```diff
@@ -25,132 +20,112 @@
...
-        res = self._sample_index[self._rank]
-        self._sample_index[self._rank] += 1
+        res = self._sample_index
+        self._sample_index += 1
         return res
 
     @property
     def current_idx(self) -> int:
-        return self._sample_index[self._rank]
+        return self._sample_index
```
voegtlel, line 128: Could directly rename the variable to `current_idx` and expose it, instead of using `@property`

src/megatron/energon/wrappers/batch_dataset.py
```diff
@@ -104,9 +75,13 @@ def __init__(
         self.batcher_config = batcher_config
         self.drop_last = drop_last
         self.error_handler = error_handler
-        self._sample_index = SampleIndex(worker_config, src=self)
-        self._generator_sample_keys = [None] * max(self.worker_config.num_workers, 1)
-        self._generator_offsets = [None] * max(self.worker_config.num_workers, 1)
+
+        self.reset_state_own()
```
voegtlel, line 79: In general, I wonder if `reset_state_own` should only be called in the worker? I.e. not in the init at all. Because the state actually only exists in the worker, right?

src/megatron/energon/wrappers/group_batch_dataset.py
```diff
@@ -19,68 +19,51 @@
...
 T_batch = TypeVar("T_batch", covariant=True)
 T_batch_sample = TypeVar("T_batch_sample", covariant=True)
 
 
-@dataclass_slots
-class BucketState(State):
-    key: Hashable
-    batch_size: int
-    samples: SampleBufferState
+# @dataclass_slots
```
voegtlel, line 34: Remove commented code

src/megatron/energon/wrappers/shuffle_buffer_dataset.py
```diff
@@ -51,8 +31,12 @@ def __init__(
         """Create a shuffle buffer for the dataset."""
         super().__init__(dataset, worker_config=worker_config)
         self.size = size
+        self.reset_state_own()
+
+    def reset_state_own(self) -> None:
         self._worker_rng = WorkerRng(self.worker_config)
-        self._active_buffer = SavableSampleBuffer(dataset, worker_config=worker_config)
+        # TODO: Former reset was via restore_state(None). Is this ok too?
```
voegtlel, line 38: I guess it would be nice to have this pulled through every class, that the savables are mostly instantiated in the init? And the respective savable will call `.reset_state_*` on the inner types?

voegtlel (changes requested):

src/megatron/energon/flavors/base_dataset.py
```diff
@@ -281,48 +244,72 @@ class SavableDataset(IterableDataset[T_sample], Generic[T_sample], ABC):
 
     worker_config: WorkerConfig
 
+    #: List of names of the fields that are saved and restored in the state.
+    _savable_fields: ClassVar[List[str]] = []
```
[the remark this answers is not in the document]
philipp-fischer: I looked up the convention an double underscore is for builtin stuff. This is just "protected" or "private".
I can make it a tuple

src/megatron/energon/flavors/base_dataset.py
```diff
@@ -281,48 +244,72 @@ class SavableDataset(IterableDataset[T_sample], Generic[T_sample], ABC):
...
+                getattr(self, key).restore_state(value)
+            else:
+                setattr(self, key, value)
+
+    @abstractmethod
+    def reset_state_own(self) -> None:
+        """Resets the state of the dataset to the initial state. Can only be called in a worker process."""
         ...
 
+    def reset_state_deep(self) -> None:
```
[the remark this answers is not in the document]
philipp-fischer: Resetting the own state *must* be overridden by the derived classes, but recursion is handled by the base class

src/megatron/energon/savable_loader.py
philipp-fischer: Yeah, but may be dangerous to remove if some legacy user code relies on it.

src/megatron/energon/wrappers/base.py
philipp-fischer: The recursion ends when we don't hit any `BaseWrapperDataset` anymore. A normal savable dataset won't have a method `reset_state_deep`, because it does not contain any inner datasets.

src/megatron/energon/wrappers/base.py
philipp-fischer: Done

src/megatron/energon/wrappers/group_batch_dataset.py
philipp-fischer: Done

src/megatron/energon/flavors/webdataset/joined_webdataset.py
philipp-fischer: You spotted a bug

src/megatron/energon/flavors/webdataset/sample_loader.py
philipp-fischer: I fixed a bug around this code, but the naming is still the same. 
We have `self.workers_slice_offsets` which is set in the main process for all workers, and we have 
```
#: The offsets of the slice slices to iterate over for the current worker
slice_offsets: Optional[Sequence[int]]
```

src/megatron/energon/wrappers/base.py
```diff
@@ -25,133 +20,123 @@
...
 
     def assert_can_restore(self) -> None:
-        self.dataset.assert_can_restore()
+        for ds in self.datasets:
+            ds.assert_can_restore()
+
+    def worker_has_samples(self) -> bool:
+        return any(ds.worker_has_samples() for ds in self.datasets)
+
+    def find_wrapped_dataset(self, cls: Type[SavableDataset]) -> Optional[SavableDataset]:
```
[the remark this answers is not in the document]
philipp-fischer: Should not break at blend actually. It should go into the inner datasets

src/megatron/energon/savable_loader.py
```diff
@@ -925,6 +851,22 @@ def _worker_command(self, *cmd_args) -> List[Any]:
...
+    def get_batch_size(self) -> Optional[int]:
+        """Try to infer micro batch size from the dataset"""
+        if isinstance(self.dataset, SavableDatasetWrapper):
+            dataset = self.dataset.dataset
+        else:
+            dataset = self.dataset
+
+        if (
+            isinstance(dataset, BaseWrapperDataset)
+            and (bds := dataset.find_wrapped_dataset(BatchDataset)) is not None
```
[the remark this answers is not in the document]
philipp-fischer: Yes, I checked it. Starting from Python 3.8

src/megatron/energon/flavors/base_dataset.py
```diff
@@ -281,48 +244,76 @@ class SavableDataset(IterableDataset[T_sample], Generic[T_sample], ABC):
...
+        for key in self._savable_fields:
+            attr = getattr(self, key)
+            if isinstance(attr, Savable):
+                state[key] = attr.save_state()
+            else:
+                # Check if this field is a simple python type or a user class
 
-        Args:
-            states: The states of the workers to merge.
+                if attr is not None and getattr(attr, "__module__", "builtins") != "builtins":
```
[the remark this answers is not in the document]
philipp-fischer: I didn't see a need. We don't seem to have a use case for that right now.
Do you think we need to check for `Savable` recursively?

src/megatron/energon/savable_loader.py
```diff
@@ -288,8 +210,10 @@ class SavableDatasetWrapper(IterableDataset[Tuple[int, int, T]], Generic[T]):
     _sample_index: int = 0
     _worker_offset: int = 0
     _last_checkpoints: List[SavableCheckpoint]
-    _restore_from: Optional[Any] = None
-    _skip_samples: List[int]
+
+    _workers_restore_from: List[Optional[SavableDatasetState]] = []
```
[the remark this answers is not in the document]
philipp-fischer: Good catch.

src/megatron/energon/flavors/base_dataset.py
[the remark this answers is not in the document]
voegtlel: But this is not for the `Savable`, but for builtins? Savable will check recursively itself?

src/megatron/energon/wrappers/base.py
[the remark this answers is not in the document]
voegtlel: And then it would probably return the first found. That would be wrong, because all inner datasets of blend are equal, thus it should return all in parallel, which is not at all needed here. So I still suggest to move this to where it's needed to avoid errors. Also, just using the first doesn't necessarily work.

src/megatron/energon/savable_loader.py
```diff
@@ -288,8 +210,10 @@ class SavableDatasetWrapper(IterableDataset[Tuple[int, int, T]], Generic[T]):
...
+    _workers_dataset_state: List[Optional[FlexState]]
```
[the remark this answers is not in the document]
voegtlel: Not doing this now.

src/megatron/energon/wrappers/base.py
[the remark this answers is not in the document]
voegtlel: Making the method private now. It has it's usefulness for debugging, but shouldn't be exposed.

voegtlel (approved):

src/megatron/energon/savable_loader.py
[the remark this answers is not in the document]
philipp-fischer: Made private

PR #83, opened 2025-03-04
Add prefetch_factor arg for loader

philipp-fischer:
For some use cases, the user needs to fine tune the `prefetch_factor` argument of the underlying pytorch dataloader. This PR allows to specify the arg when calling `get_loader` or `get_savable_loader`


voegtlel (changes requested):
The `n_checkpoints` argument is dependent of the `prefetch_factor`. We should assert that the number matches.

philipp-fischer:
> The `n_checkpoints` argument is dependent of the `prefetch_factor`. We should assert that the number matches.

Done @lvoegtle

voegtlel (approved):

PR #84, opened 2025-03-04
Add nested sample (image / video list) support for QWen-vl style of models

yaoyu-33:
This PR is raised by QWen folks. Add sample type for image / video list loading support.


philipp-fischer:
As discussed in slack, this is not the optimal approach to implement multi-image samples.

PR #88, opened 2025-03-08
fix: use EPath.open in ITarReader

shunjiad:
Use `EPath.open` to open the file with the MSC URL.


philipp-fischer:
Thanks for catching this! 
We'd somehow need a unit test with MSC object storage to not miss those.

PR #93, opened 2025-03-12
Update Audio/Video loading

voegtlel:
Changes:
* Make docs uniform with existing docs
* Fix interface to `AVData`, now using the internal stream
* Fix typing (py 3.9 compatibility)
* Updated uv lock, seemed outdated
* Unify the image tensor shapes to (channels, height, width) to avoid confusion

### DIFF
--- a/src/megatron/energon/av/decode_av_frames.py
+++ b/src/megatron/energon/av/decode_av_frames.py
@@ -132,8 +132,11 @@ class AVData:
             else:
                 audio_fps = 0
 
-            # enable multi-threaded decode for video
-            video_stream.thread_type = 3
+            # Enable multi-threaded decode for video
+            # TODO: This causes a bug which leads to a deadlock in ffmpeg when deallocating the object.
+            # Thus, disable for now.
+            # video_stream.thread_type = 3
+            video_stream.thread_type = 0
 
             # Collect metadata
             video_fps = float(video_stream.average_rate) if video_stream.average_rate else 0.0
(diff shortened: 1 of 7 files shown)


src/megatron/energon/av/decode_av_frames.py
```diff
@@ -178,15 +182,37 @@ def get_frame_batch(
 
         return video_tensor, metadata
 
+    @overload
```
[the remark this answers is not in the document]
voegtlel: Well, that's exactly the point: If you enable audio, it's none, otherwise it's a tensor. E.g. fixes typing in the tests and also in my usage in megatron

src/megatron/energon/av/decode_av_frames.py
```diff
@@ -47,7 +47,7 @@ def get_frames(
         video_decode_audio: bool = False,
         video_num_frames: int = 64,
         video_out_frame_size: tuple[int, int] = (224, 224),
-    ) -> VideoData | None:
+    ) -> Optional[VideoData]:
```
jon-barker, line 50: nit: `Optional` doesn't imply that `None` will always be returned if `VideoData` is not, so this is actually less clear.

voegtlel: Resolved: `Optional` and `| None` are equivalent: https://docs.python.org/3.10/library/typing.html#typing.Optional

jon-barker: resolved in offline discussion: VideoData | None needs python 3.10 and Optional is functionally equivalent

jon-barker:
lgtm

PR #112, opened 2025-04-08
Fastseek Fallback to Probe Mode

Queuecumber:
This PR fixes an issue with videos that are in unsupported containers or containers which have slight deviations from the expected format. 

In these cases the default Fastseek mode, container parsing, will throw an exception on an unrecognized format. Here, we catch that exception and fallback to Fastseek's probe mode which will parse the packets of the video bitstream (not the container) using ffmpeg via pyav to find keyframe information.

This is much faster than decoding the video frames but is still slower than container parsing. If the probe mode also fails, we allow the exception to propagate to the Energon user (this is almost certainly a corrupted video).

### DIFF
--- a/src/megatron/energon/av/av_decoder.py
+++ b/src/megatron/energon/av/av_decoder.py
@@ -14,7 +14,6 @@ try:
     import av
     import av.container
     import filetype
-
     from megatron.energon.av.fastseek import Fastseek
 
     AV_DECODE_AVAILABLE = True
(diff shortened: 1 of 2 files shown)


philipp-fischer:
@Queuecumber can you fix license headers and ruff? (see CI pipeline).
If you have `just` and `uv` intalled, you can run `just fix` to address the ruff linting / formatting.

src/megatron/energon/av/av_decoder.py
```diff
@@ -48,7 +48,10 @@ def __init__(self, stream: BinaryIO) -> None:
     @property
     def seeker(self) -> Fastseek:
         if self._seeker is None:
```
philipp-fischer, line 50: As discussed in slack, we might need to make sure Fastseek and pyav operating on the same stream doesn't cause any trouble, since both expect seeking at file offset 0.

philipp-fischer (changes requested):

src/megatron/energon/av/av_decoder.py
Queuecumber: Yes do you want to include that in this PR? Should be quick to add

PR #115, opened 2025-04-11
Fix packing at dataset exhaustion

### LINKED ISSUE #114: Bug: when packing is enabled self._pre_packing_buffer is not empty in some cases
When packing is enabled ; I faced a case where I got this issue that `self._pre_packing_buffer `is not empty  during `next_pre_pack` function call and `next_final_pack` calls and then it fails in `assert len(self._pre_packing_buffer) == 0`

voegtlel:
Fixes #114.

Issue occurs when dataset is exhausted, but not all packs were ejected yet.


philipp-fischer:
I verified that the test catches the issue 👍

PR #118, opened 2025-04-17
AVDecoder: Separate audio and video duration methods

philipp-fischer:
Joint audio/video duration is inaccurate, which is a problem for the last frame when extracting frames uniformly.

### DIFF
--- a/src/megatron/energon/av/av_decoder.py
+++ b/src/megatron/energon/av/av_decoder.py
@@ -479,14 +479,41 @@ class AVDecoder:
         with av.open(self.stream) as input_container:
             return len(input_container.streams.video) > 0
 
-    @overload
-    def get_duration(self, get_frame_count: Literal[True]) -> tuple[float, int]: ...
+    def get_audio_duration(self) -> Optional[float]:
+        """Get the duration of the audio stream.
+
+        Returns:
+            The duration of the audio stream in seconds
+        """
+
+        self.stream.seek(0)
+
+        with av.open(self.stream) as input_container:
+            if input_container.streams.audio:
+                audio_time_base = input_container.streams.audio[0].time_base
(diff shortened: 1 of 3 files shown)


src/megatron/energon/av/av_decoder.py
```diff
@@ -512,37 +539,21 @@ def get_duration(self, get_frame_count: bool = False) -> tuple[float, Optional[i
...
-                )
-            elif video_duration is not None:
+            if video_duration is not None and video_stream.time_base is not None:
                 duration = int(video_duration - video_start_pts) * video_stream.time_base
-            elif audio_duration is not None:
-                duration = int(audio_duration - audio_start_pts) * audio_time_base
 
             if get_frame_count and num_frames is None:
                 packets = [p for p in input_container.demux(video=0) if p.pts is not None]
                 num_frames = len(packets)
```
voegtlel, line 554:
```suggestion
                num_frames = sum(1 for p in input_container.demux(video=0) if p.pts is not None)
```
saves the list creation of length `num_frames`

src/megatron/energon/av/av_decoder.py
```diff
@@ -25,6 +25,13 @@
 from megatron.energon.av.av_data import AVData
 
 
+def initialize_av_container(input_container: av.container.InputContainer) -> None:
```
voegtlel, line 28: Move to end of file

voegtlel, line 28: Add documentation, why is this needed.
Cause 1:
Threading causes a bug which leads to a deadlock in ffmpeg when deallocating the object.
Cause 2:
Threading causes forking to break (what happens?)

voegtlel (changes requested):

PR #120, opened 2025-04-23
Fix MKV decoding time stamps

philipp-fischer:


src/megatron/energon/av/av_decoder.py
```diff
@@ -101,18 +101,18 @@ def get_video_clips(
...
+            if video_clip_ranges is not None:
+                # Convert video_clip_ranges to seeker unit
+                if video_unit == "frames" and self.seeker.unit == "pts":
+                    # Convert from frames to pts units
                     video_clip_ranges = [
                         (
-                            clip[0] / average_rate,
-                            clip[1] / average_rate,
+                            clip[0] / average_rate / time_base,
+                            clip[1] / average_rate / time_base,
```
Queuecumber, line 111: This probably should have been discussed in a previous MR but this is dangerous.

Using average rate to compute the time a given frame is displayed is error prone and its the reason why Fastseek was modified to return both the frame index and the pts for each given keyframe (otherwise you could compute the frame index from the PTS), if the notion of frame index is supported by the underlying container. 

Otherwise you should *only* be using timing information

src/megatron/energon/av/utils.py
```diff
@@ -58,12 +58,15 @@ def get_clips_uniform(
         # Special case of single frames: End point should be start of last frame
         video_fps = av_decoder.get_video_fps()
         video_spf = 1 / video_fps
-        last_start_time = total_duration - video_spf * 0.9
+        first_start_time = video_spf * 0.5
```
Queuecumber, line 61: I understand what you're doing here now, I think we need to dig into this further, it shouldn't be that complicated to do but it might be related to the problematic notion of frame timing from my last comment

tests/test_av_decoder.py
```diff
@@ -112,27 +112,33 @@ def test_decode_all_frames(self):
...
+        for video_file in ["tests/data/sync_test.mkv", "tests/data/sync_test.mp4"]:
+            print(f"================= Testing {video_file} ==================")
+            with open(video_file, "rb") as f:
+                raw_bytes = f.read()
+                stream = io.BytesIO(raw_bytes)
+
+            av_decoder = AVDecoder(stream)
+            video_duration, frame_count = av_decoder.get_video_duration(get_frame_count=True)
+
+            video_tensor = get_single_frames_uniform(
```
Queuecumber, line 124: Is this intended to test all code paths in the decoding loop?

src/megatron/energon/av/utils.py
philipp-fischer: The fix in this line was only needed for the MKV/MP4 duration difference that we saw. If we start the linspace at exactly 0, it's more sensitive to slight deviations.

philipp-fischer: We use linspace to compute `N` positions (time) in the video. Each time refers to the frame that is presented at that timestamp, i.e. it is "rounded down" to the previous frame. Hence picking time stamps in the middle of frames is "more stable"

src/megatron/energon/av/av_decoder.py
Queuecumber: It comes up a lot with "broadcast" videos (which may have originated from TV) so videos which report 29.97fps or 59.94fps and they're more common than you would think (or hope). 

These are usually stored as `30000/1001` or `60000/1001` fps and the result of that number coming out to something ridiculous is you get frames which alternate duration, for example one will have a 16ms duration and the next will have a 17ms duration, or you have have videos where the duration is expressed as constant 16ms but the presentation times have a 1ms gap. 

Guo/Cosmos ran into these regularly

Of course there's always the argument that "you don't need *exactly* the frame you were looking for" because you should only be a couple of frames off in non-pathological videos. If we go this route we should at least have a warning that the seek unit doesnt match the request unit and that it could cause inaccuracy. 

The best thing to do is if the container supports frame numbers, use frame numbers. If it only supports time, use time.

Queuecumber: BTW here is a decent explanation of where this came from: https://www.reddit.com/r/finalcutpro/comments/mm58qp/why_on_earth_do_we_have_2997fps_what_madness_is/

src/megatron/energon/av/utils.py
Queuecumber: OK I think this makes sense, the only caution I would add is that 

> picking time stamps in the middle of frames

"middle of frames" is ill-defined without duration information from each frame (from above comment)

philipp-fischer: I added a warning in the code above

src/megatron/energon/av/av_decoder.py
Queuecumber: > I think we can be off by at most 1 frame?

Most commonly yes, and my argument has always been that you probably don't need *exactly* the evenly spaced frames you were looking for anyway, so I think it's fine to allow this with a warning

PR #123, opened 2025-04-29
Add More Fallback Logic to Fastseek

Queuecumber:
Certain malformed mp4 files can yield empty keyframe lists or trigger random parse errors

This PR explicitly catches the empty keyframe list issue and also falls back to probe mode if any type of exception is thrown during parsing

### DIFF
--- a/src/megatron/energon/av/av_decoder.py
+++ b/src/megatron/energon/av/av_decoder.py
@@ -14,7 +14,6 @@ try:
     import av
     import av.container
     import filetype
-
     from megatron.energon.av.fastseek import Fastseek
 
     AV_DECODE_AVAILABLE = True
@@ -661,5 +660,7 @@ def initialize_av_container(input_container: av.container.InputContainer) -> Non
 
     for stream in input_container.streams:
         cc = stream.codec_context
-        cc.thread_type = "NONE"
-        cc.thread_count = 0
+
+        if cc is not None:
+            cc.thread_type = "NONE"
+            cc.thread_count = 0
--- a/src/megatron/energon/av/av_decoder.py
+++ b/src/megatron/energon/av/av_decoder.py
@@ -14,6 +14,7 @@ try:
     import av
     import av.container
     import filetype
+
     from megatron.energon.av.fastseek import Fastseek
 
     AV_DECODE_AVAILABLE = True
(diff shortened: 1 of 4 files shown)


src/megatron/energon/av/av_decoder.py
```diff
@@ -46,7 +46,7 @@ def __init__(self, stream: BinaryIO) -> None:
 
         try:
             self.seeker = Fastseek(self.stream)
-        except ValueError:
+        except Exception:
```
philipp-fischer, line 49: Can we use a more narrow expression instead of catch-all? 
We only want to catch fastseek-errors and not filesystem / memory errors etc., right?

src/megatron/energon/av/av_decoder.py
```diff
@@ -661,5 +661,7 @@ def initialize_av_container(input_container: av.container.InputContainer) -> Non
 
     for stream in input_container.streams:
         cc = stream.codec_context
-        cc.thread_type = "NONE"
-        cc.thread_count = 0
+
+        if cc is not None:
```
philipp-fischer, line 665: Thanks for fixing.

philipp-fischer (changes requested):

src/megatron/energon/av/av_decoder.py
Queuecumber: We could try to figure out the subset of exception types that should be propagated, but if there really is a problem that `probe=True` can't fix, then it would fail on the next line and get propagated anyway, so I'm not sure if it really matters

philipp-fischer: Say for example, something is wrong with the file stream that we pass in and it cannot be read... All videos will fall back to probe and we won't notice anything, except slowness

philipp-fischer: Btw. do we need to `self.stream.seek(0)` before the second try with probe mode?

Queuecumber: > but seems probe mode opens the file differently (cf. parse_probe with av.open vs. parse_mpeg)

This is true, so we would lose bug reporting on `parse_mpeg` and `parse_mkv` if it is a fixable bug that caused the error, but it would also stop a training run

> Btw. do we need to self.stream.seek(0) before the second try with probe mode?

I think so, technically probe mode does scan the entire file, but it's possible it does something else with the file pointer, I think we can keep it in to be safe

PR #126, opened 2025-05-08
Fix Save - Restore indexing

voegtlel:
Fix: Restoring had the worker index wrong and saving a just restored state, when none or not all workers have iterated yet.

Added test for this.

Smaller fixes:
* Worker shutdown in tests is faster.
* print rng state more compact


src/megatron/energon/savable_loader.py
```diff
@@ -458,6 +459,13 @@ def get_checkpoint(self, last_sample_indexes: List[int]) -> SavableDatasetCheckp
                     state=checkpoint.state,
                     offset=sample_index - checkpoint.sample_index,
                 )
+
+        # Immediate save after restore
+        if len(self._last_checkpoints) == 0:
```
philipp-fischer, line 464: @lvoegtle: I'm wondering, What if some workers have already restored from `_workers_restore_from` and iterated and other not?

PR #128, opened 2025-05-12
<fix>:  Missing object and 3 attributes

bobo0810:


bobo0810:
```bash
  File "/xxx/megatron-energon/src/megatron/energon/wrappers/batch_dataset.py", line 141, in flush
    batch_sample = self.batcher(batch)
  File "/xxx/megatron-energon/my_task_encoder.py", line 73, in batch
    return self._batch(samples, result_type=CaptioningRawBatch, actions={"image": batch_stack, "caption": batch_list})
  File "/xxx/megatron-energon/src/megatron/energon/task_encoder/base.py", line 334, in _batch
    return result_type(**list_samples)
TypeError: CaptioningRawBatch.__init__() got an unexpected keyword argument '__restore_key__'
```

bobo0810:
> Thank you for reporting this issue. We discussed the original design a bit and came to the conclusion, that we prefer to change the base `Batch` class, which makes the docs simpler. #131

❤️

PR #131, opened 2025-05-27
Fix Batch base class

voegtlel:


src/megatron/energon/task_encoder/base.py
```diff
@@ -235,8 +240,95 @@ def get_stateless(fn: Callable[..., T_sample]) -> bool:
...
+    batch `Batch.derive_from`."""
+
+    #: Uniquely identifies each sample in the dataset.
+    __key__: list[str]
+    #: Key for restoring the sample. This is used to restore the sample from a checkpoint. It
+    # should be a (nested) tuple of strings and integers, which can be used to index the dataset.
+    __restore_key__: Tuple[Union[str, int, tuple], ...]
+
+    #: A dataset may define a subflavor to distinguish between samples of the same sample type.
+    __subflavor__: list[Optional[str]]
```
philipp-fischer, line 255: Hm, don't we want to make the list itself optional too? I know `derive_from` will always fill it, but what if the user wants a custom batch type and he doesn't use subflavors at all?
Btw. how/when do we fade out `__subflavor__`?

src/megatron/energon/wrappers/batch_dataset.py
```diff
@@ -148,29 +151,41 @@ def flush():
...
                 pass
             except SYSTEM_EXCEPTIONS:
                 raise FatalSampleError.from_sample(batch)
             except Exception as e:
                 self.error_handler(e, batch)
+                last_failures += 1
+                if last_failures > 100:
+                    raise FatalSampleError.from_sample(
+                        batch,
+                        f"BatchDataset {self.batcher} failed 100 times in a row. Likely your dataset is broken.",
```
philipp-fischer, line 185: Is there some case you hit takes makes this necessary?

src/megatron/energon/wrappers/batch_dataset.py
```diff
@@ -138,7 +138,10 @@ def __iter__(self) -> Iterator[T_batch]:
             batch.clear()
             sample_restore_keys = []
 
+        last_failures = 0
```
philipp-fischer, line 141: I don't like how the `last_failures` variable sits in-between methods. Let's move it up to the top and name it w.g. `last_flush_failures` or so.

src/megatron/energon/task_encoder/base.py
```diff
@@ -235,8 +240,95 @@ def get_stateless(fn: Callable[..., T_sample]) -> bool:
...
 @edataclass
-class Batch(PinMemoryMixin):
-    """Base class for a batch dataclass. Provides a default implementation for pinning memory."""
+class Batch(PinMemoryMixin, ExtendableDataclassMixin):
+    """Base class for a batch dataclass. Provides a default implementation for pinning memory.
+    Additionally, it provides a future safe implementation for creating an instance from another
+    batch `Batch.derive_from`."""
+
+    #: Uniquely identifies each sample in the dataset.
+    __key__: list[str]
```
philipp-fischer, line 249: This will break existing code that already derives from `Batch` and uses `__keys__`, right?

src/megatron/energon/task_encoder/base.py
voegtlel: Maybe Energon 6.0.0 is the time to fade it out? :D

src/megatron/energon/wrappers/batch_dataset.py
philipp-fischer: same for map dataset

src/megatron/energon/task_encoder/base.py
voegtlel: `__keys__` was never specified by us. If the user uses that it'll break, yes.

src/megatron/energon/task_encoder/base.py
philipp-fischer: Can still also make `__subflavors__` optional above

PR #138, opened 2025-06-03
Add AV metadata fetching

voegtlel:
Adds `AVDecoder.get_metadata`.

### DIFF
--- a/src/megatron/energon/av/av_decoder.py
+++ b/src/megatron/energon/av/av_decoder.py
@@ -5,10 +5,12 @@ import io
 import warnings
 from collections.abc import Iterator
 from fractions import Fraction
-from typing import BinaryIO, Literal, Optional, Sequence, TypedDict, Union, overload
+from typing import BinaryIO, Literal, Optional, Sequence, Union, overload
 
 import numpy as np
 import torch
+
+from megatron.energon.edataclass import edataclass
 
 # Try importing optional dependencies
 try:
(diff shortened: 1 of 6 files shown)


src/megatron/energon/av/av_decoder.py
```diff
@@ -667,6 +743,19 @@ def __call__(
             raise ValueError(f"Invalid av_decode value: {self.av_decode}")
 
 
+class AVMetadata(TypedDict, total=False):
```
pfischer-nvidia: Why is this a typed dict and not a dataclass like AVData?

src/megatron/energon/av/av_decoder.py
```diff
@@ -582,6 +582,82 @@ def get_video_duration(
 
         return float(duration) if duration is not None else None, num_frames
 
+    def get_metadata(self, get_frame_count: bool = True) -> "AVMetadata":
```
pfischer-nvidia: Let's add a unit test that verifies the result of this function with the expected values for our test video.

src/megatron/energon/av/av_decoder.py
```diff
@@ -582,6 +582,82 @@ def get_video_duration(
...
+        """
+        self.stream.seek(0)
+        with av.open(self.stream) as input_container:
+            initialize_av_container(input_container)
+
+            metadata = AVMetadata()
+
+            if input_container.streams.video:
+                video_stream = input_container.streams.video[0]
+                assert video_stream.time_base is not None
```
pfischer-nvidia: If all of the fields in AVMetaData are optional, instead of an assertion we can simply skip the code below.

pfischer-nvidia (changes requested):

philipp-fischer (approved):

PR #141, opened 2025-06-10
Fix for uppercase extensions

philipp-fischer:
Fix AV Decoder to recognize also uppercase file extensions

### DIFF
--- a/src/megatron/energon/av/av_decoder.py
+++ b/src/megatron/energon/av/av_decoder.py
@@ -432,14 +432,10 @@ class AVDecoder:
         self,
         video_decode_audio: bool = False,
     ) -> Optional[AVData]:
-        """Decode the audio/video data with the specified parameters.
+        """Decode the entire audio/video data and return an AVData object.
 
         Args:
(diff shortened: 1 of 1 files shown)


src/megatron/energon/av/av_decoder.py
```diff
@@ -432,22 +432,21 @@ def get_frames(
...
 
         Returns:
             VideoData containing the decoded frames and metadata, or None if decoding failed
             The video tensor is in the shape (frames, channels, height, width)
             The audio tensor is in the shape (channels, samples)
         """
         extension = self._get_extension()
-        if extension in ("mov", "mp4", "webm", "mkv"):
+        if extension is not None:
+            extension = extension.lower()
```
voegtlel, line 447: Shouldn't this always be lowercase if fetching the extension from the contents?

philipp-fischer: It seems that `filetype` currently has only lower case extensions, but there is no documented guarantee for this, so to be safe, I'd keep `lower()` here.

PR #142, opened 2025-06-25
Add version to CLI and prepped dataset

philipp-fischer:


src/megatron/energon/__init__.py
```diff
@@ -1,6 +1,14 @@
 # Copyright (c) 2025, NVIDIA CORPORATION.
 # SPDX-License-Identifier: BSD-3-Clause
 
+try:
+    from importlib.metadata import version
+
+    __version__ = version("megatron-energon")
```
voegtlel, line 7: Could use `from packaging.version import Version` directly here?

src/megatron/energon/__init__.py
```diff
@@ -1,6 +1,14 @@
...
+except Exception:
+    # Fallback for development mode when package is not installed
+    __version__ = "unknown"
```
voegtlel, line 10: shouldn't we fallback to the git version? I.e. try to fetch the last tag from git, and add the number of commits on top.

src/megatron/energon/flavors/webdataset/structs.py
```diff
@@ -13,7 +13,9 @@
 class WebdatasetInfo:
     """Info about a webdataset. Format for `.nv-meta/.info.yaml` or `.nv-meta/.info.json`."""
 
-    # Maps shard name to number of samples in that shard
+    #: The version of the energon library that was used to prepare the dataset
+    energon_version: Optional[str] = None
```
voegtlel, line 17: Is `None` a good default? If we compare this in the future (i.e. if dataset.energon_version > ...), it'll require a special case. Maybe we should use `packaging.version.Version` directly to support comparisons properly.

src/megatron/energon/tools/info.py
```diff
@@ -58,6 +59,7 @@ def command(
     info_config = get_dataset_info(path)
     split_config_obj = load_yaml_json(path / MAIN_FOLDER_NAME / split_config)
 
+    ds_energon_version = info_config.get("energon_version", "unknown")
```
voegtlel, line 62: Same default here

src/megatron/energon/__init__.py
voegtlel: As a fallback of the fallback, maybe we should default to `1.0.0` for datasets which don't have a version?

voegtlel (changes requested):

PR #146, opened 2025-07-01
Synchronize end of loader for repeat=False

voegtlel:
Provides two strategies for synchronizing the end of the loader for repeat=False.

### DIFF
--- a/docs/source/advanced/epochized_blending.md
+++ b/docs/source/advanced/epochized_blending.md
@@ -25,20 +25,23 @@ splits:
         split_part: val
 ```
 
-Now, the call to `get_train_dataset` requires the additional parameter `repeat=False` to interrupt iterating after one epoch:
+Now, the call to `get_train_dataset` requires the additional parameter `repeat=False` to interrupt iterating after one epoch and use the `RedistributeLoader` to synchronize the last samples when some ranks have already exhausted their data:
 
 ```py
 from megatron.energon import get_train_dataset, get_loader, WorkerConfig
+from megatron.energon.sync_end import RedistributeLoader
 
 
-loader = get_loader(get_train_dataset(
+# Set repeat=False to avoid repeating the dataset.
+# Also add RedistributeLoader to synchronize the end of rank exhaustion. Only works with initialized torch distributed.
+loader = RedistributeLoader(get_loader(get_train_dataset(
     'metadataset.yaml',
     batch_size=2,
     shuffle_buffer_size=100,
     max_samples_per_sequence=100,
     worker_config=WorkerConfig.default_worker_config(),
     repeat=False,
-))
+)))
 
 # This will now stop iterating after the datasets have been iterated (coco 5 times, coyo-train 2
 # times and coyo-val 1 times). Of course, the data is still being shuffled between all those
@@ -54,3 +57,6 @@ for batch in loader:
 
 If used as dataset for `get_val_dataset`, the `repetitions` are ignored.
 The metadataset would also work without setting `repeat=False`, but then the shuffle buffer will shuffle samples across bounderies of epochs.
+There are two available end of iteration synchronizers:
+* `RedistributeLoader`: Redistributes samples when a rank is exhausted before other ranks. On next epoch, includes the incomplete batches' samples.
+* `StopFirstLoader`: Stop as soon as the first rank is exhausted. The next epoch will iterate until the next loader stops, restarting all ranks once.
--- a/src/megatron/energon/sync_end/redistribute.py
+++ b/src/megatron/energon/sync_end/redistribute.py
@@ -1,0 +1,451 @@
+# Copyright (c) 2025, NVIDIA CORPORATION.
+# SPDX-License-Identifier: BSD-3-Clause
+from typing import Any, Dict, Generic, Iterator, Optional, Sequence, Tuple, TypeVar, Union
+
+import torch
+import torch.distributed as dist
+
+from megatron.energon.edataclass import edataclass
+from megatron.energon.flavors.base_dataset import State
+from megatron.energon.savable_loader import (
+    BasicDataLoader,
+    SavableDataLoader,
+    SavableDataLoaderState,
+)
+from megatron.energon.worker import WorkerConfig
+from megatron.energon.wrappers.base import get_sample_restore_key
+
+__all__ = ["RedistributeLoader", "RedistributeDataLoaderState"]
+
+
+T = TypeVar("T")
+
+
+@edataclass
+class RedistributeDataLoaderState(State):
+    inner_state: SavableDataLoaderState | None
+
+    exhausted_state: bool
+    overuse_count: int
+
+    next_samples_restore_keys: list[Tuple[Union[str, int, tuple], ...]] | None
+
+    def __repr__(self):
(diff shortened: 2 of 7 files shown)


docs/source/advanced/epochized_blending.md
```diff
@@ -25,20 +25,23 @@ splits:
         split_part: val
 ```
 
-Now, the call to `get_train_dataset` requires the additional parameter `repeat=False` to interrupt iterating after one epoch:
+Now, the call to `get_train_dataset` requires the additional parameter `repeat=False` to interrupt iterating after one epoch and use the `RedistributeLoader` to synchronize the last samples when some ranks have already exhausted their data:
```
philipp-fischer:
```suggestion
Now, the call to `get_train_dataset` requires the additional parameter `repeat=False` to interrupt iterating after one epoch.
You also need to use the `RedistributeLoader` which balances the last samples among ranks when some ranks are running out of data.
```

docs/source/advanced/epochized_blending.md
```diff
@@ -25,20 +25,23 @@ splits:
...
 
 ```py
 from megatron.energon import get_train_dataset, get_loader, WorkerConfig
+from megatron.energon.sync_end import RedistributeLoader
 
 
-loader = get_loader(get_train_dataset(
+# Set repeat=False to avoid repeating the dataset.
+# Also add RedistributeLoader to synchronize the end of rank exhaustion. Only works with initialized torch distributed.
+loader = RedistributeLoader(get_loader(get_train_dataset(
```
philipp-fischer, line 38: I think it would be better if this was transparent to the user, so make `get_loader` handle this internally. 
And since the choice of `RedistributeLoader` vs. `StopFirstLoader` actually changes the data that's being iterated, this choice should be made in the metadataset and not in the code, I think.
As a property of blend_epochized.

I.e. blend_epochized can either be a list as before (chooses default `RedistributeLoader`), or it can be a dict for more customization like
```
    blend_epochized:
      phase_out_behavior: stop_first_loader
      datasets:
        - repetitions: 5
            path: ./coco
            # ... Other parameters
        - repetitions: 2
            path: ./coyo
        - repetitions: 1
            path: ./coyo
            split_part: val
```

docs/source/advanced/epochized_blending.md
```diff
@@ -54,3 +57,6 @@ for batch in loader:
 
 If used as dataset for `get_val_dataset`, the `repetitions` are ignored.
 The metadataset would also work without setting `repeat=False`, but then the shuffle buffer will shuffle samples across bounderies of epochs.
+There are two available end of iteration synchronizers:
```
philipp-fischer:
```suggestion

## Sample Redistribution

There are two options for handling the end of the data iteration:
```

src/megatron/energon/sync_end/redistribute.py
```diff
@@ -0,0 +1,452 @@
...
+            if len(samples) > 0:
+                # First use pending samples from previous iteration
+                sample = samples.pop(0)
+            else:
+                try:
+                    sample = next(self._iterator)
+                except StopIteration:
+                    # print(f"[r={rank}]: StopIteration\n", end="")
+                    self.exhausted_states[rank] = self_exhausted = 1
+            dist.all_reduce(
```
philipp-fischer, line 166: We should evaluate the impact of this synchronization (which happens in every iteration!) in a real world training to see if the training speed suffers when we have many nodes and ranks.

src/megatron/energon/sync_end/redistribute.py
```diff
@@ -0,0 +1,452 @@
...
+    exhausted_state: bool
+    overuse_count: int
+
+    next_samples_restore_keys: list[Tuple[Union[str, int, tuple], ...]] | None
+
+    def __repr__(self):
+        return f"RedistributeLoaderState(inner_state={self.inner_state!r}, exhausted_state={self.exhausted_state!r}, overuse_count={self.overuse_count!r})"
+
+
+class RedistributeLoader(Generic[T]):
```
philipp-fischer, line 37: I think this will break our reproducible scaling!
It seems the whole redistribution loader will not work if we stop and resume with a different number of ranks.
See https://github.com/NVIDIA/Megatron-Energon/pull/80

That's a problem we need to discuss possible solutions for.

src/megatron/energon/sync_end/stop_first_end.py
```diff
@@ -0,0 +1,259 @@
...
+class StopFirstDataLoaderState(State):
+    inner_state: SavableDataLoaderState | None
+    iterating_from_start: bool
+    next_sample_restore_key: Tuple[Union[str, int, tuple], ...] | None
+
+    def __repr__(self):
+        return f"StopFirstDataLoaderState(inner_state={self.inner_state!r}, iterating_from_start={self.iterating_from_start!r}, next_sample_restore_key={self.next_sample_restore_key!r})"
+
+
+class StopFirstLoader(Generic[T]):
```
philipp-fischer, line 34: Same issue with reproducible scaling here? I think this one could be easier to get it working without breaking repro scaling.

docs/source/advanced/epochized_blending.md
voegtlel: I wouldn't say "You need to use", because there are other options as well (i.e. custom handling, or those two methods provided now)

docs/source/advanced/epochized_blending.md
voegtlel: Okay, we can make this be handled in `get_loader`. I personally would prefer it to be separate though, as it's a feature on top?

Regarding moving the configuration to the metadataset:
I see your point that this slightly modifies how the data is iterated, but I'd also argue:
1. So far we don't really rely on torch distributed, where this piece of code is tightly bound to it.
2. This would also disable nesting of `blend_epochized`, because you cannot nest different (or unconfigured) `phase_out_behavior`.
3. This depends on `repeat=False` and doesn't make sense if `repeat=True`, so it's based on what the user sets in the code.
4. At least for `RedistributeLoader` it should not really change the data frequency (so far the settings in the metadataset mainly focuses on data frequency / blend).
5. If we move the boundary of metadataset to include this, then we should also have gradient accumulation, seeds, batch size, handling of incomplete batches, etc. in the config. I wouldn't want that, tbh.

Thus voting for keeping this in code, not in the metadataset config.

src/megatron/energon/sync_end/redistribute.py
voegtlel: Yes, agreed. Didn't benchmark the impact of this yet.

src/megatron/energon/sync_end/redistribute.py
voegtlel: Okay, haven't thought of this tbh. Yes, let's discuss offline

src/megatron/energon/sync_end/stop_first_end.py
voegtlel: Likely easier, but it's also the less useful of the two methods 😅

docs/source/advanced/epochized_blending.md
voegtlel: Made two sentences, but without "need".

PR #147, opened 2025-07-01
Fix epochized blend dataset repeat with fraction on multiple workers

voegtlel:
This fixes the bug, that for fractional ratios, the fraction was effectively multiplied by the `n_workers` (capped at 1).
(previously used `__len__` which corresponds to `len_rank` (now separate), fixed to use `len_worker` now)

It does not fix the issue that the total sum may be off by maximum `-1 * global number of workers`

### DIFF
--- a/src/megatron/energon/wrappers/batch_dataset.py
+++ b/src/megatron/energon/wrappers/batch_dataset.py
@@ -132,7 +132,7 @@ class BatchDataset(BaseWrapperDataset[T_batch_sample, T_batch], Generic[T_batch_
             batch.clear()
             sample_restore_keys = []
 
-        def flush():
+        def flush() -> Generator[T_batch, None, None]:
             nonlocal last_batch_failures
 
             try:
--- a/src/megatron/energon/wrappers/concat_dataset.py
+++ b/src/megatron/energon/wrappers/concat_dataset.py
@@ -32,9 +32,6 @@ class ConcatDataset(BaseWrapperDataset[T_sample, T_sample], Generic[T_sample]):
     def len_worker(self, worker_idx: int | None = None) -> int:
         return sum(dataset.len_worker(worker_idx) for dataset in self.datasets)
 
-    def len_rank(self) -> int:
-        return sum(dataset.len_rank() for dataset in self.datasets)
-
     def __iter__(self) -> Iterator[T_sample]:
         for ds_idx, dataset in enumerate(self.datasets):
             for sample in dataset:
--- a/src/megatron/energon/wrappers/group_batch_dataset.py
+++ b/src/megatron/energon/wrappers/group_batch_dataset.py
@@ -123,6 +123,7 @@ class GroupBatchDataset(
         self._buckets = {}
 
     def len_worker(self, worker_idx: int | None = None) -> int:
+        # Return an upper bound. This is for sure not correct.
         return self.dataset.len_worker(worker_idx)
 
     def __iter__(self) -> Iterator[T_batch]:
--- a/tests/test_metadataset_v2.py
+++ b/tests/test_metadataset_v2.py
@@ -973,7 +973,7 @@ class TestDataset(unittest.TestCase):
         data = list(enumerate(train_loader))
 
         # Check the overall number of samples
-        # Should be 0.7*len(ds1)55 + 1.5*len(ds2)55 = 38 + 55 + 27 (floor rounding)
+        # Should be 0.7*len(ds1) + 1.5*len(ds2) = 0.7*55 + 1.5*55 = 38 + 55 + 27 (floor rounding)
         assert len(data) == 38 + 55 + 27, len(data)
 
         sample_counts = Counter([int(s[1].text[0]) for s in data])


src/megatron/energon/savable_loader.py
```diff
@@ -91,8 +91,8 @@ def reset_state_own(self) -> None:
         self._sample_index = 0
         self._state_restored = False
 
-    def __len__(self):
-        return len(self.dataset)
+    def len_worker(self, worker_idx: int | None = None) -> int:
```
philipp-fischer, line 94: So the loader will not have `__len__` anymore? It doesn't inherit from `SavableDataset`

src/megatron/energon/wrappers/batch_dataset.py
```diff
@@ -89,23 +89,12 @@ def reset_state_own(self) -> None:
...
-        if n_batches_per_worker_ceil % self.batch_size != 0 and not self.drop_last:
-            n_batches_per_worker_ceil += 1
-
-        return (
-            n_batches_per_worker_floor * (num_workers - remaining_n_sample_workers)
-            + n_batches_per_worker_ceil * remaining_n_sample_workers
-        )
+    def len_worker(self, worker_idx: int | None = None) -> int:
+        n_samples = self.dataset.len_worker(worker_idx)
+        n_batches = n_samples // self.batch_size
```
philipp-fischer, line 94: Alternate way (more clear?):
```python
if self.drop_last:
    return n_samples // self.batch_size
else:
    return ceil(n_samples / self.batch_size)
```

src/megatron/energon/wrappers/buffer.py
```diff
@@ -74,6 +74,10 @@ def flush(self) -> Tuple[List[T_sample], Tuple[Any, ...]]:
         self._restore_keys.clear()
         return buffer, restore_key
 
+    @property
+    def buffer(self) -> List[T_sample]:
```
philipp-fischer, line 78: Why do we need this? Maybe for the refactor in packing below?
Is this just to unify `list(other_list)` copy constructors to `other_list.copy()`?

src/megatron/energon/wrappers/concat_dataset.py
```diff
@@ -29,8 +29,11 @@ def __init__(
     def reset_state_own(self) -> None:
         return
 
-    def __len__(self):
-        return sum(len(dataset) for dataset in self.datasets)
+    def len_worker(self, worker_idx: int | None = None) -> int:
+        return sum(dataset.len_worker(worker_idx) for dataset in self.datasets)
+
+    def len_rank(self) -> int:
```
philipp-fischer: Will the default not work here?

src/megatron/energon/wrappers/epochize_dataset.py
```diff
@@ -97,8 +97,18 @@ def __iter__(self) -> Iterator[T_sample]:
...
-        return self.length
+    def len_worker(self, worker_idx: int | None = None) -> int:
+        if worker_idx is None:
+            self.worker_config.assert_worker()
+            worker_idx = self.worker_config.rank_worker_id()
+        if self.worker_config.num_workers <= 1:
+            assert worker_idx == 0
+            return self.length
+        else:
+            local_length = self.length // self.worker_config.num_workers
```
philipp-fischer, line 108: The same `local_length` is computed above. Why not store it as a member of the class and just read it here?

src/megatron/energon/wrappers/group_batch_dataset.py
```diff
@@ -122,9 +122,8 @@ def reset_state_own(self) -> None:
         self._batch_sample_index = SampleIndex(self.worker_config, src=self)
         self._buckets = {}
 
-    def __len__(self):
-        # Return an upper bound. This is for sure not correct.
-        return len(self.dataset)
+    def len_worker(self, worker_idx: int | None = None) -> int:
+        return self.dataset.len_worker(worker_idx)
```
philipp-fischer, line 127: Maybe keep the comment that this is just an upper bound.

src/megatron/energon/wrappers/limit_dataset.py
```diff
@@ -43,8 +43,20 @@ def __init__(
...
-    def __len__(self) -> int:
-        return min(self.length, len(self.dataset))
+    def len_worker(self, worker_idx: int | None = None) -> int:
+        if worker_idx is None:
+            self.worker_config.assert_worker()
+            worker_idx = self.worker_config.rank_worker_id()
+        if self.worker_config.num_workers <= 1:
+            return self.length
+        else:
+            local_limit = self.length // self.worker_config.num_workers
```
philipp-fischer, line 53: Same as for `local_length` in epochize

src/megatron/energon/wrappers/repeat_dataset.py
```diff
@@ -47,17 +47,22 @@ def reset_state_own(self) -> None:
...
+            return self.dataset.len_worker(worker_idx)
+        return int(self.dataset.len_worker(worker_idx) * self.repeats)
 
     def __iter__(self) -> Iterator[T_sample]:
         assert self.repeats is not None or self.dataset.worker_has_samples(), (
             "Cannot repeat empty dataset indefinitely"
         )
 
-        ds_len = len(self.dataset)
+        # TODO: There is a small difference in the total sum of samples (across ranks) * repeats
```
philipp-fischer, line 60: So the plan is that this TODO goes away once we disallow fractional repeats, right?

tests/test_metadataset_v2.py
```diff
@@ -973,7 +973,7 @@ def test_metadataset_fixed_fractional_epochs(self):
         data = list(enumerate(train_loader))
 
         # Check the overall number of samples
-        # Should be 0.7*len(ds1) + 1.5*len(ds2) = 38 + 55 + 27 (floor rounding)
+        # Should be 0.7*len(ds1)55 + 1.5*len(ds2)55 = 38 + 55 + 27 (floor rounding)
```
philipp-fischer: Strange syntax. Revert? Or:
`Should be 0.7*len(ds1) + 1.5*len(ds2) = 0.7 * 55 + 1.5 * 55 = 38 + 55 + 27 (floor rounding)`

src/megatron/energon/savable_loader.py
voegtlel: Huh, but this is only the inner wrapper for `class SimpleSavableDatasetWrapper(BaseWrapperDataset[T, Tuple[int, int, T]], Generic[T]):` which does derive from `SavableDataset`

src/megatron/energon/wrappers/batch_dataset.py
voegtlel: I'd think this is clear enough, and you don't need to handle floats in a pure integer operation. Wouldn't waste time on discussing this small unimportant detail

src/megatron/energon/wrappers/buffer.py
voegtlel: I found out, that `list(...)` internally calls `len(...)`, which raises an exception for the buffer 🙈 Thus I needed to find another way to solve this and found `.buffer` or rather `.buffer.copy()` more clear, wouldn't you agree?

src/megatron/energon/wrappers/epochize_dataset.py
voegtlel: Because it's `worker_idx` dependent. Also, this is not a very complicated/expensive computation only happening once at the beginning 🤷

src/megatron/energon/wrappers/limit_dataset.py
voegtlel: Same answer as epochize

src/megatron/energon/wrappers/repeat_dataset.py
voegtlel: Well, either we disallow fractional repeats (then this test will be gone ;) ), or we fix the behavior, which is slightly complicated.

src/megatron/energon/savable_loader.py
philipp-fischer: You're right

src/megatron/energon/wrappers/batch_dataset.py
philipp-fischer: Well we are already discussing it :D But if you really prefer the `n_samples % self.batch_size != 0` formulation, it's ok for me... I think the other one is more readable

src/megatron/energon/wrappers/epochize_dataset.py
philipp-fischer: Makes sense. I overlooked it's dynamic depending on worker_idx

src/megatron/energon/wrappers/repeat_dataset.py
philipp-fischer: But then the title of the PR doesn't make sense, does it? I still think it needs a real description of what this PR does.

PR #148, opened 2025-07-01
Use Per-Frame Durations instead of Average 

Queuecumber:
In https://github.com/PyAV-Org/PyAV/pull/1880 I added support for the ffmpeg `duration` field on frames (it was previously only present in packets). That PR was shipped in PyAV 14.4.0 and this PR updates energon to use this field instead of a computed average duration. 

Since variable frame-rate videos can have frames with differing durations this PR makes our decode loop more accurate.


philipp-fischer:
Cool 😎

PR #150, opened 2025-07-03
Fix length warning for DataLoader. Copy small fixes from feature/new_tracing

voegtlel:
* Fix length warning for DataLoader (if iterating more data than `len()` reports)
* Fix for fork hooks
* Small typing fixes

### DIFF
--- a/src/megatron/energon/fork_hook.py
+++ b/src/megatron/energon/fork_hook.py
@@ -1,51 +1,55 @@
 # Copyright (c) 2025, NVIDIA CORPORATION.
 # SPDX-License-Identifier: BSD-3-Clause
 
-import functools
 import os
 import weakref
-from typing import Callable, Protocol, Type, TypeVar
+from dataclasses import dataclass
+from typing import Callable
 
-_after_in_child_fork_hooks = weakref.WeakKeyDictionary()
-_after_in_parent_fork_hooks = weakref.WeakKeyDictionary()
-_before_fork_hooks = weakref.WeakKeyDictionary()
-
-
-T = TypeVar("T", bound=Callable[[], None])
+_after_in_child_fork_hooks = dict()
+_after_in_parent_fork_hooks = dict()
+_before_fork_hooks = dict()
 
 
 def before_fork_hook(callable: Callable[[], None]):
     """
     Run function before the fork of a worker process. The function must be persistent.
     """
-    # Make sure, that callable is a method of object
-    assert getattr(callable, "__self__", None) is None, (
-        f"Callable must not be a method: {callable.__name__}"
-    )
-    # print(f"Adding before_fork_hook for {callable.__name__}\n", end="")
-    _before_fork_hooks[callable] = callable
+    if getattr(callable, "__self__", None):
+        self = callable.__self__
+        _before_fork_hooks[id(self)] = callable
+        weakref.finalize(self, lambda: _before_fork_hooks.pop(id(self)))
+    else:
+        _before_fork_hooks[id(callable)] = callable
+        weakref.finalize(callable, lambda: _before_fork_hooks.pop(id(callable)))
 
 
-def after_in_parent_fork_hook(callable: T):
+def after_in_parent_fork_hook(callable: Callable[[], None]):
     """
     Run function after the fork of a worker process. The function must be persistent.
     """
     # print(f"Adding after_in_child_fork_hook for {callable.__name__}\n", end="")
-    assert getattr(callable, "__self__", None) is None, (
-        f"Callable must not be a method: {callable.__name__}"
-    )
-    _after_in_parent_fork_hooks[callable] = callable
+    if getattr(callable, "__self__", None):
+        self = callable.__self__
+        _after_in_parent_fork_hooks[id(self)] = callable
+        weakref.finalize(self, lambda: _after_in_parent_fork_hooks.pop(id(self)))
+    else:
+        _after_in_parent_fork_hooks[id(callable)] = callable
+        weakref.finalize(callable, lambda: _after_in_parent_fork_hooks.pop(id(callable)))
 
 
-def after_in_child_fork_hook(callable: T):
+def after_in_child_fork_hook(callable: Callable[[], None]):
     """
     Run function after the fork of a worker process. The function must be persistent.
     """
     # print(f"Adding after_in_child_fork_hook for {callable.__name__}\n", end="")
-    assert getattr(callable, "__self__", None) is None, (
-        f"Callable must not be a method: {callable.__name__}"
-    )
-    _after_in_child_fork_hooks[callable] = callable
+    if getattr(callable, "__self__", None):
+        self = callable.__self__
+        _after_in_child_fork_hooks[id(self)] = callable
+        weakref.finalize(self, lambda: _after_in_child_fork_hooks.pop(id(self)))
+    else:
+        _after_in_child_fork_hooks[id(callable)] = callable
+        weakref.finalize(callable, lambda: _after_in_child_fork_hooks.pop(id(callable)))
 
 
 class ForkMixin:
@@ -55,18 +59,21 @@ class ForkMixin:
 
     def __init__(self, *args, **kwargs):
         super().__init__(*args, **kwargs)
+        self.__post_init__()
+
+    def __post_init__(self):
         if getattr(self.__before_fork__, "__func__", None) is not ForkMixin.__before_fork__:
-            _before_fork_hooks[self] = "__before_fork__"
+            before_fork_hook(self.__before_fork__)
         if (
             getattr(self.__after_in_child_fork__, "__func__", None)
             is not ForkMixin.__after_in_child_fork__
         ):
-            _after_in_child_fork_hooks[self] = "__after_in_child_fork__"
+            after_in_child_fork_hook(self.__after_in_child_fork__)
         if (
             getattr(self.__after_in_parent_fork__, "__func__", None)
             is not ForkMixin.__after_in_parent_fork__
         ):
-            _after_in_parent_fork_hooks[self] = "__after_in_parent_fork__"
+            after_in_parent_fork_hook(self.__after_in_parent_fork__)
 
     def __after_in_child_fork__(self):
         """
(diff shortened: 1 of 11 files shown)


src/megatron/energon/savable_loader.py
```diff
@@ -805,97 +813,80 @@ def next_id() -> int:
...
+    def __len__(self):
+        # We override this, because otherwise we'll see warnings
+        return self.dataset.len_rank()
 
-            def __next__(self):
-                try:
-                    worker_id, sample_idx, sample = next(self._iterator)
-                    outerself._worker_sample_counters[worker_id] = sample_idx
+    def __iter__(self):
+        def _inner_generator(iterator):
```
voegtlel: Note: This is a simplification from the original code using an inner class. Now this directly uses a generator which is a bit simpler imho (same below)

src/megatron/energon/savable_loader.py
```diff
@@ -942,14 +933,14 @@ def save_state_rank(self) -> Optional[SavableDataLoaderState]:
             assert isinstance(self.dataset, SimpleSavableDatasetWrapper)
             worker_states = [self.dataset.save_state()]
             assert self._next_worker_id == 0
-        elif self._persistent_iterator is None:
+        elif self._has_workers:
```
voegtlel, line 937: This is now indicated by `_has_workers` (i.e. if an iterator was created)

src/megatron/energon/savable_loader.py
```diff
@@ -1304,85 +1298,71 @@ def __init__(
...
-                    outerself.worker_config.worker_log(
-                        {
-                            "t": "BasicDataLoader.iter",
-                            "r": outerself.worker_config.rank,
-                            "w": None,
-                            "id": outerself.id,
-                            "iter_id": self.id,
-                        }
-                    )
+    def __len__(self):
```
voegtlel, line 1302: Override the original implementation to use the inner dataset's length. Avoids the "iterated more than len" warning

src/megatron/energon/watchdog.py
```diff
@@ -277,23 +277,14 @@ def watch_iter(self, iterable: Iterable[T]) -> Iterator[T]:
         Returns:
             An iterator that wraps the input iterable and watches for timeouts.
         """
-
-        it = iter(iterable)
-        while True:
-            # Automatically resets timer
+        try:
```
voegtlel, line 280: Simplification. Otherwise, `try: ... finally: if hasattr(iter, 'close'): iter.close()` was missing. This avoids the need

src/megatron/energon/wrappers/batch_dataset.py
```diff
@@ -162,6 +163,7 @@ def flush() -> Generator[T_batch, None, None]:
                     last_batch_failures = 0
                     set_sample_restore_key(batch_sample, sample_idx, *sample_restore_keys, src=self)
                     yield batch_sample
+                sample_restore_keys.clear()
```
voegtlel: This apparently was a bug before.

src/megatron/energon/fork_hook.py
```diff
@@ -87,97 +94,73 @@ def __before_fork__(self):
...
-    A protocol that defines a method that runs before and after the fork of a worker process.
+    A mixin that runs a method after the fork of a worker process.
     """
 
+    def __post_init__(self):
+        if getattr(self.__before_fork__, "__func__", None) is not ForkMixin.__before_fork__:
+            before_fork_hook(self.__before_fork__)
+        if (
+            getattr(self.__after_in_child_fork__, "__func__", None)
+            is not ForkMixin.__after_in_child_fork__
```
philipp-fischer: This is a bug (referring to `ForkMixin` instead of `DataclassForkMixin`) but the class is not used anyway, so we can remove it.

src/megatron/energon/savable_loader.py
```diff
@@ -94,6 +94,11 @@ def reset_state_own(self) -> None:
     def len_worker(self, worker_idx: int | None = None) -> int:
         return self.dataset.len_worker(worker_idx)
 
+    @property
+    def __len__(self):
+        # Note: This disables hasattr(self, "__len__"), because that attr will
+        raise AttributeError("Disabled direct length access to avoid DataLoader warnings.")
```
philipp-fischer, line 100: Isn't the length ok now? In the new version it should redirect to len_rank(). Ok maybe that's not what people expect.

src/megatron/energon/savable_loader.py
```diff
@@ -805,97 +813,80 @@ def next_id() -> int:
...
-
-                # self._debugf = open(
-                #     f"worker_samples_rank{outerself.worker_config.rank:02}_t{int(time.time())}.log", "w"
-                # )
-
-            def __iter__(self):
-                return self
+    def __len__(self):
+        # We override this, because otherwise we'll see warnings
+        return self.dataset.len_rank()
```
philipp-fischer, line 818: So for the loader it's ok that it's len_rank? I don't get it.

src/megatron/energon/savable_loader.py
philipp-fischer: Do I get it right, that this is a purely stylistic change without effect on the behavior?

src/megatron/energon/savable_loader.py
```diff
@@ -805,97 +813,80 @@ def next_id() -> int:
...
+                            "id": self.id,
+                            "iter_id": self.id,
+                        }
+                    )
 
         if self.num_workers > 0:
             # Always keep same iterator alive, as long as it yields data
-            if self._persistent_iterator is None or self._persistent_iterator.finished:
-                self._persistent_iterator = InnerIterator(super().__iter__())
+            if self._persistent_iterator is None:
```
philipp-fischer: Why can we drop the condition on `finished`?

src/megatron/energon/watchdog.py
```diff
@@ -277,23 +277,14 @@ def watch_iter(self, iterable: Iterable[T]) -> Iterator[T]:
...
-            except:
-                # On any error, disable watchdog before re-raising
-                self.disable()
-                raise
-            else:
+            for item in iterable:
                 self.disable()
                 yield item
+                self.enable()
+        finally:
```
philipp-fischer, line 286: I think this is not equivalent. Look at previous code "On any error, disable watchdog before re-raising"

src/megatron/energon/wrappers/batch_dataset.py
```diff
@@ -34,7 +35,7 @@ class BatchDataset(BaseWrapperDataset[T_batch_sample, T_batch], Generic[T_batch_
     batch_size: int
     batcher: Callable[[List[T_batch_sample]], T_batch]
     drop_last: bool
-    error_handler: Callable[[Exception, list[T_batch_sample], list[SourceInfo]], None]
+    error_handler: Callable[[Exception, List[T_batch_sample], Sequence[SourceInfo]], None]
```
philipp-fischer: Why go back to old-style `List` instead of `list`?

src/megatron/energon/wrappers/epochize_dataset.py
```diff
@@ -85,6 +85,9 @@ def __iter__(self) -> Iterator[T_sample]:
                     break
                 yield sample
 
+        if self._offset >= local_length:
```
philipp-fischer: How could this ever happen, given the modulo above?

src/megatron/energon/wrappers/packing_dataset.py
```diff
@@ -61,7 +62,7 @@ class PackingDataset(
     #: The samples are stored sequentially in the pre_packing_buffer because
     #: SavableSampleBuffer doesn't support nesting. But to keep the groups
     #: separate, we need to store the lengths of the groups here.
-    _pre_packing_lengths: List[List[int]]
+    _pre_packing_lengths: List[int]
```
philipp-fischer, line 65: Why is this even a savable field? It's never being changed. Bug?

philipp-fischer (changes requested):

src/megatron/energon/wrappers/packing_dataset.py
voegtlel: I don't see this being a savable field?

src/megatron/energon/watchdog.py
voegtlel: I don't get it. What is not equivalent? This will disable the wd before the error is raised?

src/megatron/energon/savable_loader.py
voegtlel: Yes, correct. Imho, this makes things clearer and easier to read, and easier for the tracing to be integrated 😇

src/megatron/energon/fork_hook.py
voegtlel: It's true that this is currently not being used. I can make the worker config use it instead of the crutch there which checks if the logfile needs recreation because the pid changed. (That's where it's being used in the tracing)

src/megatron/energon/savable_loader.py
voegtlel: Well, this just ensures that the DataLoader doesn't call it internally. Outside of the `DataLoader` len is still available (i.e. `len(dataloader)`). Also, nobody should access this class instances directly outside of the `SavableDataLoader`

src/megatron/energon/savable_loader.py
voegtlel: As a reference, the default implementation of `__len__` of the torch `DataLoader`:
```py
    def __len__(self) -> int:
        if self._dataset_kind == _DatasetKind.Iterable:
            # NOTE [ IterableDataset and __len__ ]
            #
            # For `IterableDataset`, `__len__` could be inaccurate when one naively
            # does multi-processing data loading, since the samples will be duplicated.
            # However, no real use case should be actually using that behavior, so
            # it should count as a user error. We should generally trust user
            # code to do the proper thing (e.g., configure each replica differently
            # in `__iter__`), and give us the correct `__len__` if they choose to
            # implement it (this will still throw if the dataset does not implement
            # a `__len__`).
            #
            # To provide a further warning, we track if `__len__` was called on the
            # `DataLoader`, save the returned value in `self._len_called`, and warn
            # if the iterator ends up yielding more than this number of samples.

            # Cannot statically verify that dataset is Sized
            length = self._IterableDataset_len_called = len(self.dataset)  # type: ignore[assignment, arg-type]
            if (
                self.batch_size is not None
            ):  # IterableDataset doesn't allow custom sampler or batch_sampler
                from math import ceil

                if self.drop_last:
                    length = length // self.batch_size
                else:
                    length = ceil(length / self.batch_size)
            return length
        else:
            return len(self._index_sampler)
```
For us, only the `Kind.Iterable` is relevant. And there the issue is that the `self._IterableDataset_len_called` must remain `None`, which it will if we override `__len__`. Then torch doesn't check if the length is in range:
```py
            if (
                self._dataset_kind == _DatasetKind.Iterable
                and self._IterableDataset_len_called is not None
                and self._num_yielded > self._IterableDataset_len_called
            ):
                warn_msg = (
                    f"Length of IterableDataset {self._dataset} was reported to be {self._IterableDataset_len_called}"
                    f"(when accessing len(dataloader)), but {self._num_yielded} samples have been fetched. "
                )
                if self._num_workers > 0:
                    warn_msg += (
                        "For multiprocessing data-loading, this could be caused by not properly configuring the "
                        "IterableDataset replica at each worker. Please see "
                        "https://pytorch.org/docs/stable/data.html#torch.utils.data.IterableDataset for examples."
                    )
                warnings.warn(warn_msg)

```

src/megatron/energon/savable_loader.py
voegtlel: Because now the generator itself sets the reference to `None` (see line 866) and additionally we have `self._has_workers`, which is more clear imho. The local generator must anyways be restarted when it was exhausted, so this is necessary.

src/megatron/energon/wrappers/packing_dataset.py
philipp-fischer: Line 79?

src/megatron/energon/watchdog.py
philipp-fischer: Ok, yes agreed. This is much cleaner.

src/megatron/energon/fork_hook.py
philipp-fischer: I would suggest to remove it and bring it back if we really use it.

src/megatron/energon/wrappers/batch_dataset.py
philipp-fischer: I don't think so. It's handled in `finally:` already.

src/megatron/energon/savable_loader.py
philipp-fischer: Ok I get that this helps prevent the torch warning. Just wondering if the len_rank is what the user expects when they call `len(dataloader)`

src/megatron/energon/savable_loader.py
philipp-fischer: I find the construct very strange that an object is instantiated in an outer scope that holds the reference to it, but then the inner object destructs the outer reference that points to itself. This is very confusing and I think we need to find a more comprehensible implementation. To be honest I didn't see anything wrong with the previous implementation.

src/megatron/energon/wrappers/packing_dataset.py
philipp-fischer: I guess I was wrong. It is being changed in `__iter__`

src/megatron/energon/savable_loader.py
voegtlel: It's what they should expect and what torch returns at the moment. `len(loader)` should return the (approx) number of samples yielded by the object, which is the `len_rank`.

src/megatron/energon/savable_loader.py
voegtlel: I find the generator easier to understand than the inner class with `__next__`.

src/megatron/energon/fork_hook.py
voegtlel: I would prefer to keep this specifically, because otherwise we need to redo it when we need it.

philipp-fischer: I'm ok with keeping it

philipp-fischer:
I think my [previous comment](https://github.com/NVIDIA/Megatron-Energon/pull/150#discussion_r2191565897) still holds.

src/megatron/energon/savable_loader.py
voegtlel: I moved the method into the class directly. Imho, this simplifies it even more and maybe it's also ok for you this way?

src/megatron/energon/savable_loader.py
```diff
@@ -817,76 +817,77 @@ def __len__(self):
...
+                                "global_idx": self._global_sample_idx,
+                            },
+                            **({} if keys is None else {"keys": keys}),
+                        }
+                    )
+                self._sample_idx += 1
+                self._global_sample_idx += 1
+                iter_idx += 1
+                yield sample
+            self._epoch_iterator = None
```
philipp-fischer, line 866: It's still a bit weird, since this is still the outer reference to itself.. But ok. It seems we are going to drop this code with the new loader.

Issue #13, opened 2024-09-24
Sample Range for Metadataset


lvoegtle:
This feature could offer multiple use-cases:
* Limiting the training dataset to a small subset for debugging
* Post-splitting a dataset into val/train/test
* Changing sampling weights for a subset

This would be implemented like:
```yaml
__module__: megatron.energon
__class__: Metadataset
splits:
  train:
    datasets:
      - sample_range: [0, 1000]
        path: ...
      - ...
```

Setting the range of the dataset to consider to the first 1k sample. Sample indexes are computed by concatenating the webdataset shards (from the `split.yaml`).

This feature only works for the innermost level, i.e. not for nested metadatasets (counting for that is not clear, because nested datasets would need to be concatenated, which is not the case for training).

Currently, for `get_val_dataset`, there already exists a `limit` option, which is different: That only limits the iterations (i.e. batches) to that number, which gives different samples than setting the sample range (range is based on sample index, the iterations may interleave different portions of the dataset).

voegtlel:
There also was a request to allow splitting the dataset by samples for train/val/test (not only by whole shards).

PR #151, opened 2025-07-04
Dataset subsets

voegtlel:
Fixes https://github.com/NVIDIA/Megatron-Energon/issues/13

Implements the sample range like this:

Example from tests:

ds1: Values [0, 55]
ds2: Values [100, 155]
ds3: Values [200, 255]

`metadataset_ratio2.yaml`:
```yaml
__module__: megatron.energon
__class__: MetadatasetV2
splits:
  train:
    # take [10, 30] from ds1, [20, 40] from ds2 and then only [20%, 80%]
    # I.e. sample range: [14, 26], 2 * [124, 136]
    subset: {range: [20%, 80%]}
    blend_epochized:
      - path: ds1
        subset: {range: [10, 30]}
      - repetitions: 2
        subset: {range: [20, 40]}
        path: ds2
```

```yaml
__module__: megatron.energon
__class__: MetadatasetV2
splits:
  train:
    subset_ratio: [0%, 50%]
    blend_epochized:
      - path: ds3
        # take [30, 50] from ds3, then first 50%, resulting in samples [230, 240]
        subset: {range: [30, 50]}
      - repetitions: 2
        # Inner sample range: [14, 26], 2 * [124, 136], total=12*3=36
        # Applying subset ratio 25%-75%: [17, 23], 2*[127, 133], total=3*6=18
        # Applying outer 50%: [17, 20], 2*[127, 130], total=3*3=9
        # Applying repetition: 2*[17, 20], 4*[127, 130], total=2*9=18
        subset: {range: [25%, 75%]}
        path: metadataset_ratio2.yaml
```

Recursively applies the `subset` (innermost first). `subset` with absolute values can only be applied to a "leaf" dataset (i.e. not a recursive metadataset).

This is 

Creating as draft up for discussion.


docs/source/advanced/subsets.md
```diff
@@ -0,0 +1,75 @@
...
+Each element can be either
+
+* a **percentage** string (e.g. `"0%"`, `"12.5%"`, `"100%"`) – interpreted relative to each inner
+  dataset size, or
+* an **absolute** integer – interpreted as a sample index. Absolute indices are only allowed for
+  *leaf* datasets (`path` to a prepared dataset containing `.nv-meta`).
+
+## Basic example
+
+The snippet below keeps the first 80 % of *COYO* for training while evaluating on the remaining
```
philipp-fischer: It's not clear to me if 100% refers to the full coyo dataset or the internal `train` split of the dataset (acc. to `split.yaml`).

docs/source/advanced/subsets.md
```diff
@@ -0,0 +1,75 @@
...
+
+Example: the outer subset `[0%, 50%]` followed by an inner subset `[25%, 75%]` results in the final
+range `[25%, 50%]` of the original dataset.
+
+Absolute indices short-circuit merging: they can **only** be specified at the leaf level and must
+not be combined with another absolute range farther up the hierarchy.
+
+## Absolute ranges
+
+Absolute indices are handy when exact sample counts are required. The following configuration takes
```
philipp-fischer: This paragraph does not mention or explain the 0 to 50% range which is also part of the example below. I think we should add that.

src/megatron/energon/flavors/webdataset/structs.py
```diff
@@ -58,6 +58,24 @@ class FilteredSample(TypedDict):
     __sources__: tuple[SourceInfo, ...]
 
 
+@edataclass
+class DatasetSubset:
+    """A subset of a dataset.
+    The range is a tuple of two values, where the first value is the start of the subset and the second value is the end of the subset.
+    The range can be either a percentage or an absolute sample index value.
```
philipp-fischer: I thought range is always percentage and absolute_range is always sample indices.
And actually range is not percentage but rather a ratio in `[0, 1]` so percentage divided by 100.

We also need an explanation here on how `range` and `absolute_range` interact. At first I thought only one of them can be set, but that's not true! So a thorough explanation is needed here.

src/megatron/energon/metadataset/metadataset_v2.py
```diff
@@ -64,8 +67,95 @@ def get_file_store(self) -> FileStore:
         return SystemFileStore(self.fs_path)
 
 
+class PercentageOrAbsoluteConverter(Converter[float | int]):
+    """Converter for percentage values or absolute values."""
```
philipp-fischer:
```suggestion
    """Converter for percentage values or absolute values.
    Integers will be returned as-is. Percentage numbers like `75%` will be converted to float ratios like `0.75`.
    """
```

EDIT: See comment below, I think we should drop the converter.

src/megatron/energon/metadataset/metadataset_v2.py
```diff
@@ -64,8 +67,95 @@ def get_file_store(self) -> FileStore:
...
+            return float(json_obj.removesuffix("%")) / 100.0
+        except (ValueError, AssertionError):
+            raise JsonValueError(
+                f"Invalid percentage value: {json_obj}", float, json_obj, path, stage
+            )
+
+    def to_json(self, obj: float | int) -> str | int:
+        if isinstance(obj, int):
+            return obj
+        assert 0 <= obj <= 100, "Percentage must be between 0 and 100"
```
philipp-fischer: This should be the inverse to `from_json`, so we expect a ratio in `[0,1]`

EDIT: See comment below, I think we should drop the converter.

src/megatron/energon/metadataset/metadataset_v2.py
```diff
@@ -64,8 +67,95 @@ def get_file_store(self) -> FileStore:
...
+        return f"{obj:.2f}%"
+
+
+@edataclass
+class Subset:
+    """
+    A subset of a dataset.
+    The range is a tuple of two values, where the first value is the start of the subset and the second value is the end of the subset.
+    The range can be either a percentage or an absolute sample index value.
```
philipp-fischer:
```suggestion
    The range can be either a ratio or an absolute sample index value.
```

src/megatron/energon/metadataset/metadataset_v2.py
```diff
@@ -64,8 +67,95 @@ def get_file_store(self) -> FileStore:
...
+    # Only internally, extracted from the range.
+    absolute_range: tuple[int, int | None] | None = dataclasses.field(default=None, init=False)
+
+    def __post_init__(self) -> None:
+        if isinstance(self.range[0], int):
+            if isinstance(self.range[1], int):
+                self.absolute_range = (self.range[0], self.range[1])
+                self.range = (0.0, 1.0)
+            else:
+                self.absolute_range = (self.range[0], None)
```
philipp-fischer: This is a weird case I don't understand. This would happen if the user enters something like `[23, 85%]`.
I think this should be forbidden. We should allow only for just an absolute range with start and end or a percentage range with start and end.

src/megatron/energon/metadataset/metadataset_v2.py
```diff
@@ -64,8 +67,95 @@ def get_file_store(self) -> FileStore:
...
+class Subset:
+    """
+    A subset of a dataset.
+    The range is a tuple of two values, where the first value is the start of the subset and the second value is the end of the subset.
+    The range can be either a percentage or an absolute sample index value.
+    The absolute range can only be used for leaf datasets.
+    """
+
+    range: tuple[
+        Annotated[float | int, PercentageOrAbsoluteConverter()],
```
philipp-fischer: This took me 5-10 minutes to understand where and why the conversion happens. Now I understand that the typed converter interprets this annotation to execute the converter. This is very obscured and will make it incredibly hard for energon dev newcomers. 
Suggestion: Don't interpret the values during the typed conversion, but later in post_init. Some of the interpretation is already there.

src/megatron/energon/flavors/webdataset/sharder.py
```diff
@@ -272,13 +274,50 @@ def _clean_offsets(offsets: Sequence[int]) -> Sequence[int]:
             int(offsets[-1]),
         )
 
+    @staticmethod
+    def _compute_subset(
```
philipp-fischer, line 278: The method `_compute_subset()` should be part of the DatasetSubset class as `apply()`.
That would also keep the logic of `absolute_range` and `range` internal to that class. This is a very specific logic which implies that absolute_range is always applied first and then range, which is a combination from nested ranges. Keeping that inside `DatasetSubset` would be better.

src/megatron/energon/typed_converter.py
```diff
@@ -227,6 +254,11 @@ class MyNamedTuple(NamedTuple):
             The input data as `inst_type`.
         """
         type_name = getattr(inst_type, "__name__", repr(inst_type))
+        if typing.get_origin(inst_type) is Annotated:
```
philipp-fischer: Please, please let's not use `Annotated` for custom init logic. Too obscure.

src/megatron/energon/metadataset/metadataset_v2.py
voegtlel: Explanation: Initially, this was part of `__post_init__`, but it was very hard to implement the follow up functions, if the typing has to be `tuple[str | float | int, str | float | int]`. Thus I ended up with this solution as the best solution. Adding a comment, explaining how the Converter works.

docs/source/advanced/subsets.md
```diff
@@ -47,9 +47,14 @@ not be combined with another absolute range farther up the hierarchy.
...
-the first **1000** samples from *COCO* and mixes them with the full *COYO* dataset using
-weight-based blending:
+Absolute indices are handy when exact sample counts are required.
+
+## Advanced example
+
+The following configuration combines the absolute ranges with the nested rules. The inner subset takes
+the first **1000** samples from *COCO* train split and mixes them with the full *COYO* train split using
+weight-based blending. The outer nesting then reduces the inner range to the first 50%, thus only
+taking the first **500** samples, mixed with the first **50%** of the dataset effectively.
```
philipp-fischer: Suggestions:
Old: 
`The outer nesting then reduces the inner range to the first 50%, thus only taking the first **500** samples, mixed with the first **50%** of the dataset effectively.`
New:
`The outer nesting then reduces the inner ranges to the first 50%, thus only taking the first **500** samples of *COCO*, mixed with the first **50%** of the *COYO* dataset effectively.`

src/megatron/energon/flavors/webdataset/structs.py
```diff
@@ -61,14 +61,53 @@ class FilteredSample(TypedDict):
 @edataclass
 class DatasetSubset:
     """A subset of a dataset.
-    The range is a tuple of two values, where the first value is the start of the subset and the second value is the end of the subset.
-    The range can be either a percentage or an absolute sample index value.
-    The absolute range can only be used for leaf datasets.
+    A range is a tuple of two values, where the first value is the start of the subset and the second value is the end of the subset.
+
+    The sharder uses the (absolute/relative) ranges to compute the subsets:
+     * `absolute_range` is applied first on the (e.g. train/val/test) subset
```
philipp-fischer: Should add that the unit is in `samples`

src/megatron/energon/flavors/webdataset/structs.py
```diff
@@ -61,14 +61,53 @@ class FilteredSample(TypedDict):
...
+     * then `range` is applied on the subset that is left
```
philipp-fischer: Should add that unit is `ratio` in `[0,1]`

src/megatron/energon/metadataset/metadataset_v2.py
```diff
@@ -86,19 +88,22 @@ def from_json(self, json_obj: Any, path: str, stage: tuple[int, ...]) -> float |
...
+        return f"{obj * 100:.2f}%"
 
 
 @edataclass
 class Subset:
     """
     A subset of a dataset.
     The range is a tuple of two values, where the first value is the start of the subset and the second value is the end of the subset.
-    The range can be either a percentage or an absolute sample index value.
+    The range can be either a ratio in percentage or an absolute sample index value.
```
philipp-fischer: But the ratio is not "in percentage"? It's in `[0,1]`

src/megatron/energon/metadataset/metadataset_v2.py
```diff
@@ -108,15 +113,15 @@ class Subset:
...
-            if isinstance(self.range[1], int):
-                self.absolute_range = (self.range[0], self.range[1])
-                self.range = (0.0, 1.0)
-            else:
-                self.absolute_range = (self.range[0], None)
-                self.range = (0.0, self.range[1])
-        elif isinstance(self.range[1], int):
-            self.absolute_range = (0, self.range[1])
-            self.range = (self.range[0], 1.0)
+            assert isinstance(self.range[1], int) or self.range[1] is None, (
```
philipp-fischer: None is not really clear. It's not documented either.

philipp-fischer (changes requested):

src/megatron/energon/metadataset/metadataset_v2.py
philipp-fischer: Out of curiosity, I just tried exactly that and it seems to work? What am I missing?

PR #159, opened 2025-07-31
Fix failure threshold

voegtlel:
1. Default failure tolerance wasn't used properly through `@stateless` without setting it.
2. it didn't work for small datasets with repetitions. Fixed that through a variable that lives in class space (but is not part of the state). TODO: Make this a state variable? (might be needed for proper state management)
3. The failure counters weren't resetted in the packing dataset

### DIFF
--- a/src/megatron/energon/task_encoder/base.py
+++ b/src/megatron/energon/task_encoder/base.py
@@ -21,7 +21,6 @@ from typing import (
     Type,
     TypeVar,
     Union,
-    cast,
     overload,
 )
 
@@ -122,7 +121,9 @@ P = ParamSpec("P")
 
 
 @overload
-def stateless(*, restore_seeds: bool = False, failure_tolerance: Optional[int] = None) -> Callable[[Callable[P, T]], Callable[P, T]]: ...
+def stateless(
+    *, restore_seeds: bool = False, failure_tolerance: Optional[int] = None
+) -> Callable[[Callable[P, T]], Callable[P, T]]: ...
 
 
 @overload
--- a/src/megatron/energon/wrappers/map_dataset.py
+++ b/src/megatron/energon/wrappers/map_dataset.py
@@ -161,11 +161,10 @@ class MapDataset(BaseWrapperDataset[T_sample, T_sample_out], Generic[T_sample, T
             except Exception as e:
                 self.error_handler(e, sample)
                 self._last_map_failures += 1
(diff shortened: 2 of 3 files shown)


src/megatron/energon/task_encoder/base.py
```diff
@@ -120,8 +121,11 @@ def batch_list(batch: List[Any]) -> Any:
 P = ParamSpec("P")
 
 
+use_default = cast(int, object())
+
+
 @overload
-def stateless(*, restore_seeds: bool = False) -> Callable[[Callable[P, T]], Callable[P, T]]: ...
+def stateless(*, restore_seeds: bool = False, failure_tolerance: Optional[int] = use_default) -> Callable[[Callable[P, T]], Callable[P, T]]: ...
```
philipp-fischer: Why this fancy casting of a random empty object? Why not use `None`?

src/megatron/energon/wrappers/batch_dataset.py
```diff
@@ -38,6 +38,7 @@ class BatchDataset(BaseWrapperDataset[T_batch_sample, T_batch], Generic[T_batch_
     _sample_index: SampleIndex
     _generator_sample_keys: Optional[Any]
     _generator_offset: Optional[int]
+    _last_batch_failures: int = 0
 
     _savable_fields = ("_sample_index", "_generator_sample_keys", "_generator_offset")
```
philipp-fischer, line 43: Just wondering: in the description of the PR you say we should make it a state variable. But why exactly? I don't think we need to restore it.

src/megatron/energon/task_encoder/base.py
voegtlel: Because `None` means "disable".

src/megatron/energon/wrappers/batch_dataset.py
voegtlel: Exactly. It's a state variable which is not restored. Still it's "state"

philipp-fischer: I'm confused, the state is defined to be the set of variables in `_savable_fields`, see https://github.com/NVIDIA/Megatron-Energon/blob/feature/failure_threshold_fix/src/megatron/energon/flavors/base_dataset.py#L292
Why would we put it there if we don't intend to restore it?

voegtlel: But it is not to be saved. It's also not config. Thus it's a "state" of the class. Without being defined in the savable fields (i.e. a third type of instance variables).

PR #160, opened 2025-07-31
Feature/jsonl

voegtlel:
Creating a draft here (i.e. code is running, docs not written yet) to allow using (indexed) jsonl directly.

Example (also check out tests):
`dataset.jsonl`:
```jsonl
{"txt": "Hello"}
{"txt": "42"}
{"txt": "I love Energon"}
```

```sh
energon prepare dataset.jsonl
```
This will create the index. That's it.

Usage (or in a metadataset v2):
```py
@stateless()
def my_cooker(sample: CrudeSample) -> TextSample:
    return TextSample(
        **basic_sample_keys(sample), text=sample["json"]["txt"]
    )


class SimpleCookingTaskEncoder(DefaultTaskEncoder):
    cookers = [Cooker(cook=my_cooker)]

train_dataset = get_train_dataset(
    "dataset.jsonl",
    ...,
    task_encoder=SimpleCookingTaskEncoder(),
)
```


src/megatron/energon/cache/file_store.py
```diff
@@ -87,3 +88,10 @@ def __init__(
 
     def get_path(self) -> str:
         return str(self.base_path)
+
+
+class JsonlFileStore(IJsonlReader, FileStore[bytes]):
+    """This dataset will directly read files from the dataset tar files from a prepared energon dataset."""
```
philipp-fischer: Dcostring wrong

src/megatron/energon/dataset_config.py
```diff
@@ -81,24 +86,44 @@ def get_dataset_from_config(
         The instantiated dataset
     """
     path = EPath(path)
-    if not check_dataset_info_present(path):
-        raise ValueError(
-            f"Path {path} does not contain a {MAIN_FOLDER_NAME}/.info.yaml or .info.json file. Did you forget to "
-            f"prepare the dataset? Please check the documentation for an introduction to dataset "
-            f"preparation."
+    dataset: BaseCoreDatasetFactory[T_sample]
+    if path.name.endswith(".jsonl"):
```
philipp-fischer: Docstring needs update where it says `Path to the folder where the .nv-meta folder is contained.`
Also `Gets a dataset from a config path.` is not really correct and needs extension.

src/megatron/energon/dataset_config.py
```diff
@@ -81,24 +86,44 @@ def get_dataset_from_config(
...
+        assert sample_type is CrudeSample or sample_type is None, (
+            f"Sample type must be CrudeSample for jsonl datasets, but got {sample_type}"
+        )
+        assert dataset_config == "dataset.yaml", (
```
philipp-fischer: This is weird, since there is no dataset.yaml at all. Should we allow `None` for the arg?
If not, at least change the error message like `Cannot specify any custom dataset config for jsonl datasets`.
Same below for `split.yaml`

src/megatron/energon/flavors/jsonl/crude_jsonl_dataset.py
```diff
@@ -0,0 +1,229 @@
...
+
+
+class CrudeJsonlDatasetFactory(
+    BaseCoreDatasetFactory[CrudeSample],
+    JsonlPreparator,
+    Sharder,
+    ErrorHandler,
+):
+    """
+    Factory class for a jsonl file as a crude dataset.
```
philipp-fischer: Can we add more explanation? What object is the factory producing? Where is it used?

src/megatron/energon/flavors/jsonl/crude_jsonl_dataset.py
```diff
@@ -0,0 +1,229 @@
...
+        part_filter: Optional[Callable[[str], bool]] = None,
+        handler: Callable[
+            [Exception, Optional[str], Optional[list[SourceInfo]]], None
+        ] = reraise_exception,
+    ):
+        """
+        Factory for a jsonl file as a crude dataset.
+
+        Args:
+            path: Path to the dataset.
```
philipp-fischer:
```suggestion
            path: Path to the jsonl dataset file.
```

src/megatron/energon/flavors/jsonl/crude_jsonl_dataset.py
```diff
@@ -0,0 +1,229 @@
...
+            as_split_part: The split part to use for the jsonl file.
+            split_part: Which part to load (e.g. 'train', 'val', 'test'). If it does not match the as_split_part, the dataset will be empty.
```
philipp-fischer: I think `split_part` needs to be removed while `as_split_part` makes sense.

src/megatron/energon/flavors/jsonl/crude_jsonl_dataset.py
```diff
@@ -0,0 +1,229 @@
...
+                How many epochs to shuffle over if training.
+                If = 1, every sample is seen exactly once per epoch.
+                If > 1, samples (or rather shard slices) are shuffled within this number of epochs
+                (i.e. randomly selected without replacement).
+                If -1, the shards are effectively shuffle over infinite epochs (i.e. shard slices
+                are drawn with replacement).
+            parallel_shard_iters: Number of parallel opened shards per worker, shuffling between.
+            max_samples_per_sequence: Maximum number of samples per sequence (=how many samples
+                    will be sequentially iterated).
+            split_config: Config file to use for shard split definitions.
```
philipp-fischer: split_config does not exist

src/megatron/energon/flavors/jsonl/ijsonl.py
```diff
@@ -0,0 +1,296 @@
...
+                    json.loads(data)
+    """
+
+    def __init__(self, fileobj: BinaryIO):
+        self.fileobj = fileobj
+
+    def seek(self, offset: int):
+        self.fileobj.seek(offset)
+
+    def next(self, offset: int | None = None, size: int | None = None) -> bytes | None:
```
philipp-fischer, line 267: Do we ever have the case where we need the bytes directly? If not, I'd suggest to do the `json.loads` right here.

src/megatron/energon/flavors/jsonl/ijsonl.py
```diff
@@ -0,0 +1,296 @@
...
+
+
+@edataclass
+class CacheEntry:
+    ijsonl_index_reader: IJsonlIndexReader
+    lookahead_offset: Optional[int] = None
+    lookahead_byteoffset: Optional[int] = None
+
+
+class CachedIJsonlOffsetReader:
```
philipp-fischer, line 105: I wish we could unify this with `CachedItarOffsetReader`

src/megatron/energon/flavors/jsonl/crude_jsonl_dataset.py
```diff
@@ -0,0 +1,229 @@
...
+        return super()._load_sample(sample)
+
+    def config(self) -> Dict[str, Any]:
+        return dict(
+            **super().config(),
+            subflavors=self.subflavors,
+        )
+
+
+class DecoderCrudeJsonlDatasetFactory(DefaultCrudeJsonlDatasetFactory):
```
philipp-fischer: Can't we simplify this stack of classes? (`DecoderCrudeJsonlDatasetFactory`, `DefaultCrudeJsonlDatasetFactory`)
I know it's meant to resemble what we have with webdataset, but here it's so much simpler. The only decoder we will ever need is `json.loads`. Everything else should fail.
I'd suggest having only one subclass that directly executes `json.loads` and adds the subflavor. Without offering the possibility to customize the decoder.

src/megatron/energon/flavors/webdataset/itar.py
```diff
@@ -54,8 +55,7 @@ def __iter__(self) -> Generator[int, None, None]:
             yield struct.unpack("Q", raw)[0]
 
     def __len__(self) -> int:
-        self.itar.seek(0, 2)
```
philipp-fischer, line 57: Was there an issue with this method to get the size?

src/megatron/energon/metadataset/loader.py
```diff
@@ -26,7 +26,7 @@ def load_dataset(
         )
         return mds
     path = EPath(path)
-    if path.is_file():
+    if path.is_file() and not path.name.endswith(".jsonl"):
         mds = load_config(
```
philipp-fischer, line 32: It seems we have multiple of these checks throughout the code. Can we introduce a helper function like `is_path_metadataset()` or `is_path_dataset()`. Maybe you have an even better idea.

src/megatron/energon/wrappers/batch_dataset.py
```diff
@@ -38,6 +38,7 @@ class BatchDataset(BaseWrapperDataset[T_batch_sample, T_batch], Generic[T_batch_
     _sample_index: SampleIndex
     _generator_sample_keys: Optional[Any]
     _generator_offset: Optional[int]
+    _last_batch_failures: int = 0
```
philipp-fischer, line 41: The failure stuff is a bit unrelated to jsonl. Shall we split the PR?

tests/test_jsonl_dataset.py
```diff
@@ -0,0 +1,278 @@
...
+    @staticmethod
+    def create_text_test_dataset(
+        path: Path, txt_range: Iterable[int], key_range: Iterable[int], prefix: str = ""
+    ):
+        """Creates a small dummy test dataset for testing purposes."""
+
+        # Write jsonl file
+        with open(path, "w") as wf:
+            for key, txt in zip(key_range, txt_range):
+                # Write individual files to shards
```
philipp-fischer: Comment doesn't make sense?

tests/test_jsonl_dataset.py
```diff
@@ -0,0 +1,278 @@
...
+        assert len(train_loader) == 10, f"Expected 10 samples, got {len(train_loader)}"
+
+        train_order1 = [text for _, data in zip(range(50), train_loader) for text in data.text]
+        print(train_order1[:10])
+        print(Counter(train_order1))
+        assert len(Counter(train_order1)) == 10
+        assert all(v == 5 for v in Counter(train_order1).values())
+
+
+if __name__ == "__main__":
```
philipp-fischer, line 316: Can we add one test with multiple ranks and workers? Just to see that the slicing works well.

philipp-fischer (changes requested):
Of course we still need docs for this and adapt the docs for metadataset, for dataset format on disk, and for data preparation.

src/megatron/energon/dataset_config.py
voegtlel: A Sentinel `DEFAULT` value would be better, but python still does not have good support for that 😞

src/megatron/energon/flavors/webdataset/itar.py
voegtlel: I found a strange behavior with msc. In the end it was unrelated to this. Nevertheless kept this change, because previously were doing `len(self)` in the `__init__` anyways, so we can also directly store the length and return it here instead of recomputing every time.

src/megatron/energon/wrappers/batch_dataset.py
voegtlel: Didn't I also create a separate PR for this? 🫣

src/megatron/energon/flavors/jsonl/crude_jsonl_dataset.py
voegtlel: Yeah, I found that not so nice as well. But our global interface expects something like this 🙈 If you have a good suggestion how to satisfy the global interface?

src/megatron/energon/flavors/jsonl/ijsonl.py
voegtlel: Well, the code is using the default webdataset decoding (via the `Decoder`). If we want to be compatible with that, we also have to return the bytes here.

src/megatron/energon/flavors/jsonl/ijsonl.py
voegtlel: It should be possible for the index reader/writer, but the `CachedIJsonlOffsetReader` has quite a few small changes (as the key is the index in the jsonl file, things are slightly different). I think unification would abstract stuff quite a lot and not necessarily simplify.

src/megatron/energon/flavors/jsonl/crude_jsonl_dataset.py
voegtlel: I'd like to have the interfaces compatible in this case, i.e. using the `Decoder` from the `TaskEncoder.decoder` for reading bytes -> json. Thus I'd prefer to keep this. Also, if we consider extending this to e.g. folder + `.nv-meta` + `jsonl`(s), we'd need this.

PR #161, opened 2025-08-12
Implement using the sample_parts table in the SQLite database

voegtlel:
* Implement using the sample_parts table in the SQLite database if available for faster retrieval of sample parts
* Also allows to have "invalid" webdataset files, where sample_keys are duplicated.

### DIFF
--- a/src/megatron/energon/flavors/webdataset/indexing.py
+++ b/src/megatron/energon/flavors/webdataset/indexing.py
@@ -9,7 +9,7 @@ from typing import BinaryIO, Generator, List, Optional, Tuple, Union
 from numpy import int8
 
 from megatron.energon.epathlib import EPath
-from megatron.energon.flavors.webdataset.itar import ITarSamplePointer
+from megatron.energon.flavors.webdataset.itar import ITarRawSamplePartPointer, ITarSamplePointer
 from megatron.energon.flavors.webdataset.thread_local_sqlite import ThreadLocalSqlite
 from megatron.energon.local_copy import ensure_local_copy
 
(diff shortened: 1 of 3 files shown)


src/megatron/energon/flavors/webdataset/indexing.py
```diff
@@ -252,6 +267,21 @@ def __init__(self, sqlite_path: EPath):
 
         self.db = ThreadLocalSqlite(path)
 
+    def db_version(self) -> int:
```
philipp-fischer: suggest to rename to db_has_sample_parts and return bool

src/megatron/energon/flavors/webdataset/indexing.py
```diff
@@ -327,7 +357,37 @@ def get_sample_count(self) -> int:
...
+
+        Returns:
+            Pointer to the sample part raw data.
+        """
+        assert self.db is not None, "Database is closed"
+
+        row = self.db.select_one(
+            "SELECT sp.tar_file_id, sp.content_byte_offset, sp.content_byte_size "
+            "FROM sample_parts AS sp "
+            "JOIN samples AS s "
```
philipp-fischer, line 374: I hope this is fast. The join will be executed for every single sample part that we retrieve, so multiple times per sample. We didn't have a join before, but just a linear traversal of parts of a single sample.

src/megatron/energon/flavors/webdataset/itar_reader.py
```diff
@@ -112,6 +112,38 @@ def _get_itarfile_cached(self, tar_file_id: int) -> ITarFile:
 
         return self.itar_files_cache[tar_file_id]
 
+    def _get_part_by_raw_sample_pointer(
+        self,
+        raw_sample_pointer: ITarSamplePointer,
```
philipp-fischer: We are misusing `ITarSamplePointer` here, which says that it points to a sample and not a sample part. I think we should introduce a new dataclass even if the fields are similar.

src/megatron/energon/flavors/webdataset/itar_reader.py
```diff
@@ -112,6 +112,38 @@ def _get_itarfile_cached(self, tar_file_id: int) -> ITarFile:
...
+        entry_name: str,
+    ) -> tuple[bytes, SourceInfo]:
+        """
+        Get a sample from the dataset.
```
philipp-fischer: Docstring is not accurate,

PR #162, opened 2025-08-20
Custom dataloader

voegtlel:
Implementing a custom dataloader instead of `torch.data.utils.DataLoader`:
* Better customizability required for python nogil
* Fixing broken fork shutdown logic
* Improving savability (instead of regular checkpoints with rewind, can use the actual checkpoint), calls to the worker without a crutch command thread


src/megatron/energon/wrappers/group_batch_dataset.py
```diff
@@ -72,6 +84,10 @@ class GroupBatchDataset(
     _batch_sample_index: SampleIndex
     _buckets: Dict[Hashable, Bucket[T_batch_sample]]
 
+    _savable_fields = ("_group_key_sample_index", "_batch_sample_index")
+    # Buckets are saved manually
+    _state_fields = ("_buckets",)
```
philipp-fischer: Write own class which is savable to move into `_savable_fields`

philipp-fischer (commented):
TODO: Check if we can add a check to detect users utilizing global RNG in taskencoder.

src/megatron/energon/task_encoder/base.py
```diff
@@ -975,6 +1043,15 @@ def build_val_datasets(
 
         return dataset
 
+    def reset_state(self) -> None:
```
philipp-fischer: There was a bug here, (setting self.rng although property) but no test covers it.

src/megatron/energon/dataloader/pin_memory.py
```diff
@@ -0,0 +1,102 @@
...
+    This class is used to pin memory of samples in the primary process.
+    """
+
+    def __init__(self, device: str | torch.device):
+        self._device = device
+
+    def _pin_memory(self, sample: TSample) -> TSample:
+        return PinMemoryMixin.sample_pin_memory(sample, self._device)
+
+    def __call__(self, sample: Future[TSample]) -> Future[TSample]:
```
philipp-fischer, line 30: I believe this method is never called, because instead `_pin_memory` is directly being accessed. What's the idea behind the `__call__` pattern?

src/megatron/energon/loader.py
```diff
@@ -17,38 +22,35 @@ def get_savable_loader(
...
         worker_config: Deprecated. Please pass this to the dataset instead.
-        checkpoint_every_sec: This is the time in seconds after which an internal checkpoint is
-            saved. It may take the same duration to restore a checkpoint, but introduces additional
-            overhead during reading data from the dataset, so this should be chosen accordingly.
-            Only applies if using workers.
-        checkpoint_every_min_n_samples: Overwrites the minimum number of samples between
-            checkpoints. Defaults to `number of workers * 2`. Only applies if using workers.
-        n_checkpoints: The number of internal checkpoints to keep. Only applies if using workers.
-            If None, computes a suitable value.
+        worker_type: The type of worker to use.
```
philipp-fischer: Needs much more explanation. Maybe in the docstring above. But should at least list the options

src/megatron/energon/loader.py
```diff
@@ -17,38 +22,35 @@ def get_savable_loader(
...
+        gc_freeze_at_start: If True, the garbage collector is frozen at the start of the loader.
+        gc_collect_every_n_steps: The number of steps after which the garbage collector is called.
+        prefetch_factor: The factor by which to prefetch the dataset.
         cache_pool: If set, the cache pool to use for the dataset.
         watchdog_timeout_seconds: The timeout in seconds. If None, the watchdog is disabled.
         watchdog_initial_timeout_seconds: The initial timeout in seconds. If None, the timeout is the same as watchdog_timeout_seconds.
         fail_on_timeout: If True, stops the whole process upon timeout, after printing a stack trace.
+        pin_memory: If True, the dataset is pinned to memory.
```
philipp-fischer: "The dataset is pinned to memory". That doesn't make sense?

src/megatron/energon/task_encoder/base.py
voegtlel: I think we wanted to get rid of this anyways (i.e. not making the `TaskEncoder` `Savable`)

src/megatron/energon/dataloader/pin_memory.py
voegtlel: See line 229 of dataloader.py:
```py
self._pin_memory(
    CallableFuture(functools.partial(self.restore_sample, sample_key))
)
```
and line 317:
```py
self._prefetching_samples[worker_idx].append(
    self._pin_memory(worker.prefetch_next())
)
```

PR #164, opened 2025-08-22
Implement caching out an object to a file

voegtlel:
* Implement `cache.to_cache` to move a pickleable object to the cache for later usage
* Tests are missing yet

### DIFF
--- a/src/megatron/energon/cache/file_cache_pool.py
+++ b/src/megatron/energon/cache/file_cache_pool.py
@@ -8,6 +8,7 @@ import random
 import string
 import tempfile
 import threading
+import uuid
 from concurrent.futures import Future, ThreadPoolExecutor
 from pathlib import Path
 from typing import Any, Dict, Literal, Optional, Tuple, TypeVar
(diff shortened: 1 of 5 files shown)


src/megatron/energon/cache/file_cache_pool.py
```diff
@@ -66,6 +67,44 @@ def __del__(self):
...
+        if self._data is None:
+            with open(self.cache_path, "rb") as f:
+                self._data = pickle.load(f)
+            self.cache_path.unlink()
+            self.cache_path = None
+        return self._data
+
+    def __del__(self):
+        if self.cache_path is not None:
+            self.cache_path.unlink()
```
philipp-fischer: Could add try-except around this if the cache has been cleared by someone else already

src/megatron/energon/cache/file_cache_pool.py
```diff
@@ -66,6 +67,44 @@ def __del__(self):
...
 
+@edataclass
+class CacheFileLazy(Lazy[T]):
+    """
+    Represents a reference to a cached object without deduplication.
+    """
+
+    cache_path: Path | None
+
+    _data: Optional[T] = None
```
philipp-fischer, line 80: Needs comment that explains it is set only when fetching it back

PR #171, opened 2025-10-07
feat: including prefetch_file flag in mutlistorageclient library 

ankitmaster08:
- Added support for partial file caching with configurable chunk sizes, prefetch control, and source version validation to optimize performance for large files in multi-storage-client library
- Ran benchmark [script](https://github.com/NVIDIA/Megatron-Energon/blob/feature/benchmark_all/src/megatron/energon/tools/benchmark.py) to verify improvement in performance.


shunjiad:
Can you update the MSC version to `>=0.31.0` in the `pyproject.toml` file?

PR #174, opened 2025-10-15
Add error handlers to user calls when restoring a sample

philipp-fischer:
When restoring a checkpoint with changed TaskEncoder methods (e.g. cookers), errors thrown inside those changed methods during the restore, may crash the whole loader.
This PR adds error handlers to the restore phase, just like we have them during normal data iteration.


voegtlel (changes requested):
I vote for not catching exceptions when restoring samples. We should crash with a proper exception saying, that loading is inconsistent. If a sample could be loaded (i.e. it has a restore key), it should also be reproducible.

philipp-fischer:
@lvoegtle, I don't know how it happened, but after releasing, I realized those commits were already on develop 🤔 
If we want to improve the behavior of the error handling during restore we should probably create a new PR

PR #177, opened 2025-10-31
Make AVDecoder picklable also for videos

philipp-fischer:
Previously, trying to pickle an `AVDecoder` object (for example by `to_cache`) that was created on an mpeg file,
would result in an error `AttributeError: Can't pickle local object 'parse_mpeg.<locals>.<lambda>'`

This PR fixes it and tests this case.

### DIFF
--- a/src/megatron/energon/av/fastseek/containers/mpeg.py
+++ b/src/megatron/energon/av/fastseek/containers/mpeg.py
@@ -1,6 +1,5 @@
 # Copyright (c) 2025, NVIDIA CORPORATION.
 # SPDX-License-Identifier: BSD-3-Clause
-from collections import defaultdict
 from itertools import accumulate
 from typing import Any, Generator
 
@@ -184,8 +183,8 @@ def parse_atoms(file: BitsType) -> Generator[Atom, None, None]:
 def parse_mpeg(file: BitsType) -> dict[int, SortedList]:
     sync_samples = {}
     decode_timestamps = {}
-    presentation_time_offsets = defaultdict(lambda: defaultdict(lambda: 0))
-    start_offsets = defaultdict(lambda: 0)
+    presentation_time_offsets = {}
+    start_offsets = {}
     current_track = -1
     for a in parse_atoms(file):
         if a.name == "tkhd":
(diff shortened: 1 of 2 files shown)


voegtlel (approved):

src/megatron/energon/av/fastseek/containers/mpeg.py
```diff
@@ -223,14 +222,16 @@ def parse_mpeg(file: BitsType) -> dict[int, SortedList]:
...
         for track_id in sync_samples.keys():
+            if track_id not in keyframes:
+                keyframes[track_id] = SortedList()
             for keyframe_number in sync_samples[track_id]:
                 pts = (
                     decode_timestamps[track_id][keyframe_number]
-                    + start_offsets[track_id]
-                    + presentation_time_offsets[track_id][keyframe_number]
+                    + start_offsets.get(track_id, 0)
+                    + presentation_time_offsets.get(track_id, [])[keyframe_number]
```
Queuecumber, line 234: I think this will crash for an empty `presentation_time_offsets` since `.get(...)` will return `[]` which is the indexed with `[keyframe_number]`

src/megatron/energon/av/fastseek/containers/mpeg.py
```diff
@@ -184,8 +183,8 @@ def parse_atoms(file: BitsType) -> Generator[Atom, None, None]:
 def parse_mpeg(file: BitsType) -> dict[int, SortedList]:
     sync_samples = {}
     decode_timestamps = {}
-    presentation_time_offsets = defaultdict(lambda: defaultdict(lambda: 0))
-    start_offsets = defaultdict(lambda: 0)
+    presentation_time_offsets = {}
```
Queuecumber, line 186: This might cause a crash when processing CTTS atoms, I think that is where the defaultdict would previously have been initialized for a given `track_id` key

src/megatron/energon/av/fastseek/containers/mpeg.py
Queuecumber: The logic here looks correct other than above issue (which would be fixed by `.get(keyframe_number, 0)`

PR #180, opened 2025-11-03
Feature/update dependencies

voegtlel:
Update msc to >=0.33, fix test, add new extra


pyproject.toml
```diff
@@ -25,7 +25,7 @@ dependencies = [
   "click",
   "dataslots; python_version<'3.10'",
```
philipp-fischer: Can we drop this line? We are >3.10 anyway

PR #181, opened 2025-11-03
Implement proper error handler to task encoder

voegtlel:
See docs

### DIFF
--- a/docs/source/advanced/error_handling.md
+++ b/docs/source/advanced/error_handling.md
@@ -4,7 +4,8 @@ SPDX-License-Identifier: BSD-3-Clause -->
 # Error Handling
 
 Energon provides configurable error handling to gracefully manage exceptions during data loading and processing.
-By default, errors are logged and the sample is skipped, allowing training to continue.
+By default, errors in task encoder code are logged and the sample is skipped, allowing training to continue.
+Some system exceptions (clearly programming errors) always cause immediate failure.
 
 By default, if more than 100 samples in a row raise an error, processing will also fail. This is configurable for the
 {py:class}`~megatron.energon.task_encoder.base.TaskEncoder` user-functions by setting the `@stateless(failure_tolerance=100)` parameter.
@@ -30,6 +31,12 @@ Default is {py:func}`log_exception <megatron.energon.log_exception>` which logs 
 
 **`restore_error_handler`:** Handles exceptions during checkpoint restoration.
 Default is {py:func}`reraise_exception <megatron.energon.reraise_exception>` which fails fast to ensure checkpoint integrity.
+
+## System Exceptions
+
+Programming errors and critical system issues always cause immediate failure and are never handled by error handlers:
+`SystemError`, `SyntaxError`, `ImportError`, `StopIteration`, `StopAsyncIteration`, `MemoryError`, `RecursionError`, `ReferenceError`, `NameError`, `UnboundLocalError`, and {py:exc}`FatalSampleError <megatron.energon.FatalSampleError>`.
+{py:exc}`FatalSampleError <megatron.energon.FatalSampleError>` is raised automatically when consecutive failure tolerance is exceeded or when a system exception occurs during sample processing.
 
 ## Built-in Error Handlers
 
@@ -123,9 +130,9 @@ Set `failure_tolerance=0` to disable tolerance checking for a specific function.
 Tolerance limits count *consecutive* failures. A single successful sample resets the counter.
 ```
 
-## Skip Samples Explicitly
+## Skip or Fail Explicitly
 
-Raise {py:exc}`SkipSample <megatron.energon.SkipSample>` to explicitly skip a sample without logging an error. E.g. for skipping specific errors:
+Raise {py:exc}`SkipSample <megatron.energon.SkipSample>` to explicitly skip a sample without logging an error:
 
 ```python
 from megatron.energon import SkipSample
@@ -135,5 +142,18 @@ def process_sample(sample):
         ...
     except MySpecificError:
         raise SkipSample()
+    return sample
+```
+
+Raise {py:exc}`FatalSampleError <megatron.energon.FatalSampleError>` to cause immediate failure, bypassing the error handler:
+
+```python
+from megatron.energon import FatalSampleError
+
+def process_sample(sample):
+    try:
+        ...
+    except MyFatalError as e:
+        raise FatalSampleError.from_sample(sample, "Critical corruption detected") from e
     return sample
 ```


src/megatron/energon/sample_utils.py
```diff
@@ -0,0 +1,208 @@
...
+
+import dataclasses
+import itertools
+from typing import Any
+
+import numpy as np
+import torch
+
+
+def default_get_keys(batch: Any) -> list[str] | None:
```
philipp-fischer: 1. Suggest to include `batch` in the name
2. It seems this is used for logging. But in the places where keys are logged, don't we want to rather log the sources? The key can be ambiguous

src/megatron/energon/sample_utils.py
```diff
@@ -0,0 +1,208 @@
...
+    elif (
+        isinstance(batch, dict)
+        and "keys" in batch
+        and all(isinstance(k, str) for k in batch["keys"])
+    ):
+        return batch["keys"]
+    return None
+
+
+def format_sample_compact(
```
philipp-fischer, line 64: Nice

src/megatron/energon/sample_utils.py
```diff
@@ -0,0 +1,208 @@
...
+    elif isinstance(sample, torch.Tensor):
+        try:
+            min_val = sample.min().item()
+            max_val = sample.max().item()
+            return f"Tensor(shape={sample.shape}, dtype={sample.dtype}, device={sample.device}, min={min_val}, max={max_val})"
+        except (RuntimeError, ValueError):
+            # Handle empty tensors or non-numeric dtypes
+            return f"Tensor(shape={sample.shape}, dtype={sample.dtype}, device={sample.device})"
+    elif isinstance(sample, np.ndarray):
+        return f"np.ndarray(shape={sample.shape}, dtype={sample.dtype})"
```
philipp-fischer: Why no min/max for numpy? Also maybe show first and last values?

src/megatron/energon/errors.py
```diff
@@ -126,3 +46,184 @@ def wrapper(*args, **kwargs):
...
+    and raising a FatalSampleError when a tolerance threshold is exceeded.
+
+    Example:
+        error_ctx = ErrorContext(
+            name="MapDataset.map_fn",
+            handler=self.worker_config.global_error_handler
+            tolerance=100,
+        )
+
+        with error_ctx.handle_errors(sample, get_source_info):
```
philipp-fischer: This doesn't match the method signature.

src/megatron/energon/errors.py
```diff
@@ -126,3 +46,184 @@ def wrapper(*args, **kwargs):
...
+            name: Name of the operation being tracked (for error messages).
+            handler: Error handler function to call on exceptions. Takes (exception, sample, sources).
+                If None, exceptions will be raised after incrementing the counter.
+            tolerance: Maximum number of consecutive failures before raising FatalSampleError.
+                Set to 0 to disable tolerance checking.
+        """
+        self.name = name
+        self.tolerance = tolerance
+        self.handler = handler
+        self._consecutive_failures = 0
```
philipp-fischer:
```suggestion
        self.reset()
```

src/megatron/energon/errors.py
```diff
@@ -126,3 +46,184 @@ def wrapper(*args, **kwargs):
...
+    Args:
+        error_handler: Function to call when an exception occurs. Takes (exception, sample, sources).
+        sample: The sample being restored.
+    """
+    try:
+        yield
+    except SkipSample:
+        # Unexpected skip sample
+        try:
+            raise ValueError(f"Unexpected skip sample {sample} during restoration.")
```
philipp-fischer: This is a new exception instance. We are losing the connection to the original code location. We should capture the `SkipSample` error with `as e_skip` above and then do `raise ... from e_skip` here in this line

src/megatron/energon/errors.py
```diff
@@ -126,3 +46,184 @@ def wrapper(*args, **kwargs):
...
+    except GeneratorExit:
+        # Unexpected skip sample
+        try:
+            raise ValueError(
+                f"Unexpected generator early stopping for sample {sample} during restoration."
+            )
+        except Exception as e:
+            error_handler(e, sample, get_source_info(sample))
+    except SYSTEM_EXCEPTIONS:
+        raise FatalSampleError.from_sample(sample)
```
philipp-fischer: Keep track of original exception here as well.
Actually I think we should do this everywhere, also in the ErrorContext class. (do `raise ... from e` instead of turning `e` into a string)

src/megatron/energon/task_encoder/base.py
```diff
@@ -464,9 +462,7 @@ def preencode_sample(
         return sample
 
     @stateless
-    def postencode_sample(
-        self, sample: T_sample
-    ) -> Union[T_encoded_sample, Generator[T_encoded_sample, None, None]]:
+    def postencode_sample(self, sample: T_sample) -> T_encoded_sample:
```
philipp-fischer, line 465: Unrelated, but ok

src/megatron/energon/task_encoder/loader.py
```diff
@@ -31,13 +32,16 @@ def _split_kwargs(kwargs: dict) -> dict:
     return loader_kwargs
 
 
-def _split_deprecated_decoder_kwargs(kwargs: dict, task_encoder: TaskEncoder) -> None:
+def _split_deprecated_decoder_kwargs(
+    kwargs: dict, task_encoder: TaskEncoder, worker_config: WorkerConfig
+) -> None:
     """
     auto_decode: bool = True,
```
philipp-fischer, line 47: I know it was broken before, but can you add a little description to the docstring, explaining why it's needed and what it does?

src/megatron/energon/task_encoder/loader.py
```diff
@@ -51,6 +55,20 @@ def _split_deprecated_decoder_kwargs(kwargs: dict, task_encoder: TaskEncoder) ->
     if "video_decode_audio" in kwargs:
         decoder_kwargs["video_decode_audio"] = kwargs.pop("video_decode_audio")
 
+    if kwargs.get("handler", None) is not None:
```
philipp-fischer: This just doesn't fit into this function, because the function specifically filters out decoding kwargs. One option would be to rename it to `_split_deprecated_dataset_kwargs`.

If we keep this line, why not in the same style as the ones above? `if "handler" in kwargs:`

src/megatron/energon/wrappers/batch_dataset.py
```diff
@@ -133,7 +127,7 @@ def __iter__(self) -> Iterator[T_batch]:
             sample_restore_keys = []
 
         def flush() -> Generator[T_batch, None, None]:
-            try:
+            with self._batch_failure_handler.handle_errors(batch):
```
philipp-fischer, line 130: I like how we can get rid of the large except blocks now. Good idea.

src/megatron/energon/wrappers/packing_dataset.py
```diff
@@ -98,9 +93,6 @@ def __init__(
         sample_encoder: Optional[Callable[[T_sample], T_encoded_sample]] = None,
         sample_encoder_stateless: bool = False,
         packer_config: Optional[Union[Dict[str, Any], Callable[[], Dict[str, Any]]]] = None,
-        error_handler: Callable[
-            [Exception, List[T_sample], Sequence[SourceInfo]], None
-        ] = log_exception,
         pre_packer_failure_tolerance: int = 100,
```
philipp-fischer, line 96: Just wondering: How would the user configure the failure tolerance? They don't have access to these args, do they?

tests/test_dataset.py
```diff
@@ -1769,6 +1773,111 @@ def test_info_captioning_dataset(self):
...
+        )
+
+        try:
+            # Iterate through the loader - errors should be handled by custom handler
+            batches = []
+            for i, batch in enumerate(loader):
+                batches.append(batch)
+                if i >= 9:  # Get 10 batches (50 samples total)
+                    break
+        except ShouldRaiseException:
```
philipp-fischer: I think there's a function called `assertRaises` or so.

src/megatron/energon/errors.py
voegtlel: Removed it entirely and set default.

src/megatron/energon/sample_utils.py
voegtlel: I only moved this code from `log_sample_dataset`. But I agree, this should be adapted. Should add a separate PR to fix this (it's more than just changing this function).

src/megatron/energon/sample_utils.py
voegtlel: (also just moved this :D)

src/megatron/energon/task_encoder/loader.py
voegtlel: I'd rather not. It's deprecated anyways, so people shouldn't use it, only if they already have it in their code?

src/megatron/energon/wrappers/packing_dataset.py
voegtlel: They do. They can set it via the `@stateless(failure_tolerance=...)`

philipp-fischer: Ah cool, I didn't know actually.

philipp-fischer (approved):

docs/source/advanced/error_handling.md
```diff
@@ -0,0 +1,139 @@
+<!--- Copyright (c) 2025, NVIDIA CORPORATION.
+SPDX-License-Identifier: BSD-3-Clause -->
+
+# Error Handling
+
+Energon provides configurable error handling to gracefully manage exceptions during data loading and processing.
+By default, errors are logged and the sample is skipped, allowing training to continue.
```
philipp-fischer: Need to differentiate here. Not all errors are ignored. I would mention that this applies only to task encoder code and not to system exceptions. We can also add a list of system exceptions below.

docs/source/advanced/error_handling.md
```diff
@@ -0,0 +1,139 @@
...
+
+By default, if more than 100 samples in a row raise an error, processing will also fail. This is configurable for the
+{py:class}`~megatron.energon.task_encoder.base.TaskEncoder` user-functions by setting the `@stateless(failure_tolerance=100)` parameter.
```
philipp-fischer, line 11: Maybe not a real issue, but just wondering: What if the user has a non-stateless function and still wants to specify the failure tolerance?

docs/source/advanced/error_handling.md
```diff
@@ -0,0 +1,139 @@
...
+```
+
+Set `failure_tolerance=0` to disable tolerance checking for a specific function.
+
+```{admonition} Note
+:class: important
+Tolerance limits count *consecutive* failures. A single successful sample resets the counter.
+```
+
+## Skip Samples Explicitly
```
philipp-fischer: Would rename to "Skip or Fail Explicitly" and also explain that the user can raise `FatalSample` error themselves to skip the error handler in fatal cases.

docs/source/advanced/error_handling.md
voegtlel: Well, if it's non-stateless, they cannot use saving. Otherwise it'll not have any effect. So it also doesn't hurt if they make it stateless :sweat_smile:

PR #182, opened 2025-11-04
Feature: Store media metadata in sqlite database next to the media

philipp-fischer:
# Media Metadata Support and Integration

## Summary

- Adds tools to extract media metadata for images, audio, and video assets and store them in a SQLite database next to the media
- Works for WebDatasets as well as filesystem datasets
- Data preparation introduces CLI flags (`--media-metadata`, `--media-filter`) to opt-in to metadata generation with customizable detection strategies (extension, type, glob pattern)
- Exposes media metadata retrieval through `FileStore` implementations (`SystemFileStore`, `WebdatasetFileStore`, `DecodeFileStore`) to access metadata from inside a cooker
- Supports filesystem datasets via a new dataset type for pure-filesystem auxiliary datasets

## Examples for Data Preparation

- Prepare a WebDataset with metadata for all media parts (determined by file extension):

  ```bash
  energon prepare /data/my_webdataset --media-metadata --media-filter EXT
  ```

- Prepare the same WebDataset but restrict metadata generation to MP4 videos only using a glob filter:

  ```bash
  energon prepare /data/my_webdataset --media-metadata --media-filter "*.mp4"
  ```

- Prepare a plain filesystem folder recursively, storing metadata inside `.nv-meta/index.sqlite`
    - In this example the media is selected by determining the file type using the binary file header (file contents)

  ```bash
  energon prepare /mnt/datasets/raw_media --media-metadata --media-filter TYPE --no-progress
  ```

## Example Use in Cooker

```python
from megatron.energon.task_encoder.cooking import cooker


@cooker()
def describe_media(sample: dict, *, media_aux):
    entry_name = sample["image_name"]
    metadata = media_aux.get_media_metadata(entry_name)
    if metadata is None:
        raise RuntimeError("Dataset was not prepared with --media-metadata")

    return TextSample(
        **basic_sample_keys(sample),
        text=str(metadata),
    )
```

## Open TODOs or Items for Discussion
- [x] Add parallel processing for indexing media
- [x] Figure out what to do with primary datasets that have duplicate keys
- [x] Currently fixed `_IMAGE_EXTENSIONS` list
- [ ] Can we process metadata for 2B files? How slow is it, what will the size of the sqlite be?
- [x] Database locking errors. And we still have DELETE mode somewhere
- [ ] Docs

### DIFF
--- a/docs/source/basic/data_prep.md
+++ b/docs/source/basic/data_prep.md
@@ -218,7 +218,7 @@ If you are preparing a dataset with media files, energon can retrieve and store 
 This information will be stored inside an SQLite database file next to the dataset.
 Later, inside the [cooker](crude-data), you can access this information using the {py:meth}`get_media_metadata <megatron.energon.FileStore.get_media_metadata>` method of the {py:class}`FileStore <megatron.energon.FileStore>`.
 
-#### During normal initial preparaion of a WebDataset
+#### During normal initial preparation of a WebDataset
 
 ```sh
 > energon prepare --media-metadata --media-by-extension /path/to/dataset


src/megatron/energon/cache/file_store.py
```diff
@@ -76,6 +85,54 @@ def get_path(self) -> str:
...
+                    "Run `energon prepare --media-metadata` to generate it."
+                )
+            return None
+        metadata_type, metadata_json = row
+        return deserialize_media_metadata(metadata_type, metadata_json)
+
+    def _ensure_media_metadata_reader(self) -> ThreadLocalSqlite:
+        assert self.base_dir is not None
+        if self._media_metadata_reader is None:
+            sqlite_path = self.base_dir / MAIN_FOLDER_NAME / "index.sqlite"
```
voegtlel: What if `base_dir` is a non-local msc uri?

src/megatron/energon/cache/file_store.py
```diff
@@ -76,6 +85,54 @@ def get_path(self) -> str:
     def __str__(self):
         return f"SystemFileStore(base_dir={self.base_dir})"
 
+    def get_media_metadata(self, key: str) -> Optional[MediaMetadataBase]:
+        if self.base_dir is None:
+            raise RuntimeError("Media metadata requires a base directory for SystemFileStore")
+
+        reader = self._ensure_media_metadata_reader()
+        entry_key = key
```
voegtlel: just a rename? o.O

src/megatron/energon/flavors/webdataset/metadata.py
```diff
@@ -158,7 +159,12 @@ def get_dataset_type(path: EPath) -> EnergonDatasetType:
             return EnergonDatasetType.METADATASET
         else:
             return EnergonDatasetType.INVALID
-    elif check_dataset_info_present(path):
-        return EnergonDatasetType.WEBDATASET
+    elif path.is_dir():
```
voegtlel: I believe `is_dir` is not optimal with some remote filesystems, it may be unefficient. I'd rather directly check for the files that we're looking for.

voegtlel (changes requested):

src/megatron/energon/cache/file_store.py
philipp-fischer: Yea, sorry I had removed some other code and forgot this.

src/megatron/energon/cli/main.py
```diff
@@ -10,7 +10,12 @@
 from megatron.energon.tools.info import command as info_command
 from megatron.energon.tools.lint import command as lint_command
 from megatron.energon.tools.mount import command as mount_command
-from megatron.energon.tools.prepare import command as prepare_command
+from megatron.energon.tools.prepare import (
+    command as prepare_command,
+)
+from megatron.energon.tools.prepare import (
+    prepare_media_command,
```
voegtlel: This is not in the style of the other commands. Although I prefer not using `command as` (i.e. prefering this one). Either change all to the right name, or keep all `command`?

src/megatron/energon/cli/main.py
```diff
@@ -10,7 +10,12 @@
 from megatron.energon.tools.info import command as info_command
 from megatron.energon.tools.lint import command as lint_command
 from megatron.energon.tools.mount import command as mount_command
-from megatron.energon.tools.prepare import command as prepare_command
+from megatron.energon.tools.prepare import (
+    command as prepare_command,
```
voegtlel: Why is this wrapping in the formatting? o.O

src/megatron/energon/flavors/webdataset/indexing.py
```diff
@@ -53,39 +66,64 @@ def __init__(self, sqlite_path: EPath):
...
-        self.db.execute(
-            """
-            CREATE TABLE samples (
-                tar_file_id INTEGER,
-                sample_key TEXT,
-                sample_index INTEGER,
-                byte_offset INTEGER,
-                byte_size INTEGER
+        if self.enable_sample_tables:
+            if self.reset_tables:
```
voegtlel: Does it make sense not to reset, but to run `sample_tables`?

src/megatron/energon/flavors/webdataset/indexing.py
```diff
@@ -140,39 +178,69 @@ def append_part(
             (tar_file_id, sample_index, part_name, content_byte_offset, content_byte_size),
         )
 
+    def append_media_metadata(
```
voegtlel, line 190: Should we also have an add batch method (e.g. collecting 1024 entries, then adding all at once)? Is this efficient enough?

src/megatron/energon/media/extractor.py
```diff
@@ -0,0 +1,215 @@
+# Copyright (c) 2025, NVIDIA CORPORATION.
+# SPDX-License-Identifier: BSD-3-Clause
+
+from __future__ import annotations
```
voegtlel: remove

tests/test_av_decoder.py
```diff
@@ -222,29 +222,29 @@ def test_video_audio_sync(self):
     def test_pickle_decoder(self):
         """Test AVDecoder on a video file can be pickled and unpickled."""
         av_decoder = AVDecoder(io.BytesIO(Path("tests/data/sync_test.mp4").read_bytes()))
-        
+
```
voegtlel, line 225: Why this diff? I thought I formatted develop already in the last commit. Also, why don't the formatting tests complain about this?

src/megatron/energon/tools/prepare.py
```diff
@@ -409,5 +446,90 @@ def progress_fn(els, length=None):
     click.echo("Done")
 
 
+@click.command(name="prepare-media")
```
voegtlel: `prepare-media-metadata`? :sweat_smile:

src/megatron/energon/tools/prepare.py
```diff
@@ -409,5 +446,90 @@ def progress_fn(els, length=None):
...
+    "--progress/--no-progress",
+    default=True,
+)
+@click.option(
+    "--num-workers",
+    type=int,
+    default=16,
+    help="Number of workers to use to scan files",
+)
+@click.option(
```
voegtlel: Why? -> Delete

src/megatron/energon/tools/prepare.py
```diff
@@ -409,5 +446,90 @@ def progress_fn(els, length=None):
...
+    "--media-metadata/--no-media-metadata",
+    default=True,
+    help="Compute and store media metadata during preparation.",
+)
+@click.option(
+    "--media-filter",
+    type=str,
+    default=None,
+    help="Media detection strategy: EXT (extension, default), TYPE (by filetype), or a glob pattern such as '*.jpg'.",
```
voegtlel: I thought we wanted to allow multiple of these? Such that the user can specify "*.jpg *.png" in one?

voegtlel: Also, I think it's not clear that `EXT`/`TYPE` will automaticaly use the default supported media types. Also, `TYPE` is unclear vs `EXT`. Isn't "type" the generic word for the type of the file? I think we detect the filetype either by extension or by content. I'd rather use `EXTENSION` and `CONTENT` or better `extension` and `content` (why capitalize) and explain, that we use standard webdataset decoder semantics for the detection of the file types in those cases.

src/megatron/energon/media/extractor.py
```diff
@@ -0,0 +1,215 @@
...
+    """Configuration for media detection during dataset preparation."""
+
+    strategy: MediaFilterStrategy
+    pattern: str | None = None
+
+    @classmethod
+    def parse(cls, value: str | None) -> "MediaFilterConfig":
+        if value is None:
+            value = "EXT"
+
```
voegtlel:
```suggestion
    def parse(cls, value: str = "EXT") -> "MediaFilterConfig":

```
Also, see comment on naming "EXT"

src/megatron/energon/media/extractor.py
```diff
@@ -0,0 +1,215 @@
...
+        if value_upper == MediaFilterStrategy.EXT.value:
+            return cls(strategy=MediaFilterStrategy.EXT)
+
+        if value_upper == MediaFilterStrategy.TYPE.value:
+            return cls(strategy=MediaFilterStrategy.TYPE)
+
+        return cls(strategy=MediaFilterStrategy.PATTERN, pattern=value)
+
+
+_IMAGE_EXTENSIONS: set[str] = {
```
voegtlel, line 121: Shouldn't this already exist in the default Webdataset Decoder? I believe that should be shared?

src/megatron/energon/tools/prepare.py
```diff
@@ -409,5 +446,90 @@ def progress_fn(els, length=None):
...
+)
+@click.option(
+    "--media-metadata/--no-media-metadata",
+    default=True,
+    help="Compute and store media metadata during preparation.",
+)
+@click.option(
+    "--media-filter",
+    type=str,
+    default=None,
```
voegtlel: Instead of `default=None` rather make the default visible here.

tests/test_crudedataset.py
```diff
@@ -341,6 +420,57 @@ def create_crude_text_test_dataset(path: Path, offset: int):
                 )
             )
 
+    @staticmethod
+    def create_multimedia_webdataset(path: Path):
```
voegtlel, line 424: Nice

tests/test_media_pipeline.py
```diff
@@ -0,0 +1,133 @@
...
+from megatron.energon.media.metadata import (
+    AVMetadata,
+    ImageMetadata,
+    MediaMetadataType,
+    deserialize_media_metadata,
+    serialize_media_metadata,
+)
+
+
+class MediaPipelineTests(unittest.TestCase):
```
voegtlel: This whole thing is obviously ai generated, the style is completely different from the other tests. I'd rather remove this whole class. There is a proper whole-system test in test_crudedataset, that covers this stuff already. So let's not maintain this test as well.

voegtlel (changes requested):

src/megatron/energon/tools/prepare.py
philipp-fischer: Potentially want to add more media preparation options to this command in the future

src/megatron/energon/media/extractor.py
philipp-fischer: Yea, that's why I had the TODO item in the PR description. I looked at WebDataset, and for image extensions, they have a list on top of this file: https://github.com/webdataset/webdataset/blob/main/src/webdataset/autodecode.py
But for video and audio, the extension lists are hidden inside the decoder handlers. So maybe we should keep our own lists?

tests/test_av_decoder.py
philipp-fischer: I don't know, but I think it's correct not to have spaces there.

src/megatron/energon/flavors/webdataset/indexing.py
philipp-fischer: That is not specific to media, that would be for all samples I guess? So maybe not in this PR, but in general I like the idea.

src/megatron/energon/flavors/webdataset/tar_patcher.py
```diff
@@ -0,0 +1,533 @@
...
+        raise click.ClickException(f"In-place rename not supported: {issues}")
+
+    click.echo("OK: in-place modification is possible under current rules.")
+
+    if dry_run:
+        return
+
+    click.echo("Applying prefix in-place...")
+    try:
+        patcher.apply_prefix(tar_path, prefix, scan_result=scan_result)
```
voegtlel: Parameter scan_result does not exist.

src/megatron/energon/flavors/webdataset/tar_patcher.py
```diff
@@ -0,0 +1,533 @@
...
+        with tqdm(
+            total=len(tasks),
+            desc="Scanning dataset",
+            unit="shards",
+            disable=not self._show_progress,
+        ) as dataset_pbar:
+            with ProcessPoolExecutor(max_workers=max_workers) as executor:
+                for tar_file, result in executor.map(
+                    _scan_tar_worker, tasks, chunksize=(len(tasks) // max_workers) or 1
+                ):
```
voegtlel: Note: I replaced both executor usages to be unordered, i.e. futures returns as they complete, not blocking in the order they were submitted.

src/megatron/energon/cache/base.py
```diff
@@ -167,3 +167,32 @@ def decode(self, fname: str, data: bytes) -> Any:
...
+    def get(self, key: str, sample: Any = None) -> Any:
+        if key.startswith("."):
+            key = f"{self._current_key}{key}"
+        return self._inner.get(key, sample)
+
+    def get_path(self) -> str:
+        return self._inner.get_path()
+
+    def get_media_metadata(self, key: str) -> Optional[MediaMetadataBase]:
+        if key.startswith("."):
```
philipp-fischer, line 230: Good idea, since the key can never start with a dot

src/megatron/energon/task_encoder/base.py
```diff
@@ -701,46 +688,36 @@ def build_cook_crude_sample(
...
-                    # Make the exception throw through for the sample being loaded
-                    raise SystemError("Error getting primary auxiliary dataset") from e
-            return primary_aux
-
-        if aux is not None:
-            cook_fn = functools.partial(
-                self.cook_crude_sample, get_primary_aux=_get_primary_aux, **aux
-            )
+        for cooker in self.cookers:
+            if cooker.is_match(subflavors):
```
philipp-fischer, line 695: Do I get it right, that you did this so the matching cooker is only searched once and not on every sample?

voegtlel: exactly

voegtlel:
Will do docs in a separate pr. Need this merged.

PR #209, opened 2026-02-26
Add support for DSS URLs

philipp-fischer:
Make EPath and metadatasets support paths like this:
```
__module__: megatron.energon
__class__: MetadatasetV2
splits:
  train:
    blend:
      - path: dss://charts123_zh@v0
        weight: 1
        aux:
          media_source: dss://charts2500@v0
```

### DIFF
--- a/src/megatron/energon/epathlib/epath.py
+++ b/src/megatron/energon/epathlib/epath.py
@@ -6,7 +6,16 @@ import re
 import shutil
 from pathlib import Path as PathlibPath
 from pathlib import PurePosixPath
-from typing import BinaryIO, Generator, Literal, Optional, TextIO, Tuple, Union, overload
+from typing import (
+    BinaryIO,
+    Generator,
+    Literal,
+    Optional,
+    TextIO,
+    Tuple,
+    Union,
+    overload,
+)
 
 import multistorageclient as msc
 
@@ -130,11 +139,14 @@ class EPath:
 
     @staticmethod
     def _split_protocol(path: str) -> Tuple[Optional[str], Optional[str], str]:
-        regex = re.compile(r"^(?P<protocol>[a-z]+)://(?P<profile>[^/]+?)/(?P<path>.+)$")
+        regex = re.compile(r"^(?P<protocol>[a-z]+)://(?P<profile>[^/]+?)(?:/(?P<path>.*))?$")
         m = regex.match(path)
         if m is None:
             return None, None, path
-        return m.group("protocol"), m.group("profile"), m.group("path")
+        inner_path = m.group("path")
+        if not inner_path:
+            inner_path = ""
+        return m.group("protocol"), m.group("profile"), inner_path
 
     @property
     def _internal_str_path(self) -> str:
--- a/src/megatron/energon/epathlib/epath.py
+++ b/src/megatron/energon/epathlib/epath.py
@@ -237,13 +237,20 @@ class EPath:
 
     @property
     def url(self) -> str:
-        if self.is_local():
+        if self.is_pure_local():
             return self._internal_str_path
         int_path_str = str(self.internal_path)
+        if self.profile == "dss":
+            if int_path_str.startswith("/"):
+                int_path_str = int_path_str[1:]
+            return f"dss://{int_path_str}"
         return f"msc://{self.profile}{int_path_str}"
 
-    def is_local(self) -> bool:
+    def is_pure_local(self) -> bool:
         return self.profile == DEFAULT_PROFILE_NAME
+
+    def is_local(self) -> bool:
+        return self.profile == DEFAULT_PROFILE_NAME or self.profile == "dss"
 
     def local_path(self) -> PathlibPath:
         if not self.is_local():
--- a/src/megatron/energon/epathlib/epath.py
+++ b/src/megatron/energon/epathlib/epath.py
@@ -250,12 +250,9 @@ class EPath:
 
     def is_local(self) -> bool:
         if self.profile == "dss":
-            try:
-                # Tries to get metadata, which will fail if the local file or folder does not exist
-                self.fs.info(self._internal_str_path)
-                return True
-            except FileNotFoundError:
-                return False
+            # For now, a DSS path is always considered local.
+            # Note that this does not mean it exists on the local filesystem.
+            return True
         else:
             return self.profile == DEFAULT_PROFILE_NAME
 
--- a/src/megatron/energon/epathlib/epath.py
+++ b/src/megatron/energon/epathlib/epath.py
@@ -1,6 +1,7 @@
 # Copyright (c) 2025, NVIDIA CORPORATION.
 # SPDX-License-Identifier: BSD-3-Clause
 
+import os
 import re
 import shutil
 from pathlib import Path as PathlibPath
@@ -31,7 +32,9 @@ class EPath:
     for more information.
     """
 
-    # The path without the protocol. Can also be in S3 for example
+    __slots__ = ("internal_path", "profile", "fs")
+
+    # The path without the protocol/profile. Can also be in S3 for example
     internal_path: PurePosixPath
     # The profile used to access the file system
     profile: str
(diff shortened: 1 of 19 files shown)


src/megatron/energon/epathlib/epath.py
```diff
@@ -189,14 +236,29 @@ def parent(self) -> "EPath":
...
 
-    def is_local(self) -> bool:
+    def is_pure_local(self) -> bool:
         return self.profile == DEFAULT_PROFILE_NAME
 
+    def is_local(self) -> bool:
+        if self.profile == "dss":
+            try:
+                # Tries to get metadata, which will fail if the local file or folder does not exist
+                self.fs.info(self._internal_str_path)
```
voegtlel: It think this should be different: `is_local` should not check for existence, but should check if the file maps to a local path? I.e. if `NVDATASET_CACHE_DIR` is local.

src/megatron/energon/metadataset/loader_interface.py
```diff
@@ -46,6 +54,20 @@ def post_initialize(self, mds_path: Optional[EPath] = None):
...
+    def scan(self, mds_path: Optional[EPath] = None):
+        """Recursively discover dataset references without loading leaf datasets."""
+        raise NotImplementedError(f"{type(self).__name__} does not implement scan()")
+
+    def get_scanned_dataset_references(
+        self,
+        *,
+        split_part: Union[Literal["train", "val", "test"], str],
+    ) -> List[ScannedDatasetReference]:
+        """Flatten the leaf dataset references from a scanned metadataset."""
```
voegtlel: Docs incomplete. Add more details. This is the interface description.

src/megatron/energon/metadataset/loader_interface.py
```diff
@@ -46,6 +54,20 @@ def post_initialize(self, mds_path: Optional[EPath] = None):
         """Called to finally initialize the dataset."""
         ...
 
+    def scan(self, mds_path: Optional[EPath] = None):
+        """Recursively discover dataset references without loading leaf datasets."""
```
voegtlel: Docs incomplete

src/megatron/energon/metadataset/loader_interface.py
```diff
@@ -46,6 +54,20 @@ def post_initialize(self, mds_path: Optional[EPath] = None):
...
+        raise NotImplementedError(f"{type(self).__name__} does not implement scan()")
+
+    def get_scanned_dataset_references(
```
voegtlel: personally, I find `scanned dataset reference` quite unclear.

src/megatron/energon/metadataset/loader_interface.py
```diff
@@ -46,6 +54,20 @@ def post_initialize(self, mds_path: Optional[EPath] = None):
         """Called to finally initialize the dataset."""
         ...
 
+    def scan(self, mds_path: Optional[EPath] = None):
```
voegtlel: Returntype typing is missing.
We have `get_datasets`, `get_scanned_dataset_references` and `scan`. I find the `scan` naming out of place. Why is there the `mds_path` arg? Is this method supposed to be internal (that arg doesn't make sense externally?)?

is this supposed to be an alternative initialization to `post_initialize`? If yes, then now this class is a mix of two different initializations. If that's the case, I believe this is hard to understand and fuzzy.

Can't this and `get_scanned_dataset_references` be a single method that returns that list of flattened datasets?

tests/test_metadataset_v2.py
```diff
@@ -311,6 +311,107 @@ def test_nested_metadataset(self):
...
+        # Check that the first dataset in the inner blend
+        # points to "ds1" and that the dataset is not instantiated.
+        inner_train = nested_ref._dataset.splits["train"]
+        inner_leaf_ref = inner_train.blend[0]
+        assert isinstance(inner_leaf_ref.path, EPath)
+        assert inner_leaf_ref.path == EPath(self.dataset_path / "ds1")
+        assert inner_leaf_ref._dataset is None
+
+    def test_scan_metadataset_preserves_missing_v2_leaf_and_aux(self):
+        from megatron.energon.metadataset.metadataset_v2 import (
```
voegtlel: Why all these local imports? (this one and all the other new imports) Typically we do that on top?

tests/test_epathlib.py
```diff
@@ -279,6 +286,90 @@ def test_msc_s3(self):
             # assert not EPath("msc://s3test_msc/test").is_dir()
             assert not EPath("msc://s3test_msc/test/dir").is_dir()
 
+    def test_metadataset_v2_dss_path_parsing_str(self):
```
voegtlel, line 289: There is no actual test, which tests using a dss dataset, right? I.e. iterating samples from a dss dataset

voegtlel (changes requested):

src/megatron/energon/epathlib/epath.py
voegtlel, line 259: Need adaption for dss (this must return the `NVDATASET_CACHE_DIR / self._internal_str_path` in that case)

tests/test_epathlib.py
philipp-fischer: Yes that's right, but I think that's ok. We assume it is a normal local dataset. And the tests makes sure that the path resolution works.

src/megatron/energon/epathlib/epath.py
philipp-fischer: I think we're fine because `_internal_str_path` handles this?

voegtlel: I don't think so. That is just `self.internal_path = self._resolve(f"/{profile}/{path}")`. But here we need the resolved path to `NVDATASET_CACHE_DIR` as I just wrote. This is the place, where the resolution of the internal path of `NVDATASET_CACHE_DIR` has to happen. Otherwise `NVDATASET_CACHE_DIR` is only used as `.fs`, no other usages / effect, i.e. we would access `/mydataset@v0/path`, which is wrong.

tests/test_epathlib.py
voegtlel: I'd like to see an e2e test for dss paths if that's ok. We have that for all the other cases.

src/megatron/energon/metadataset/loader_interface.py
```diff
@@ -46,6 +62,31 @@ def post_initialize(self, mds_path: Optional[EPath] = None):
         """Called to finally initialize the dataset."""
         ...
 
+    def traverse(
```
voegtlel, line 67: Am I right to assume that you didn't want to enforce implementing this method, by omitting `@abstractmethod`?

voegtlel (changes requested):

src/megatron/energon/metadataset/metadataset.py
```diff
@@ -67,6 +76,40 @@ def post_initialize(self, mds_path: Optional[EPath] = None):
...
+                use None only for top-level metadatasets.
+            split_part: Split inherited from the parent traversal. If this reference defines its
+                own split override, that split takes precedence for nested traversal and the
+                returned leaf reference.
+
+        Returns:
+            A single leaf `TraversedDatasetReference` for direct dataset references, or the
+            flattened traversal result of the nested metadataset when this reference points to one.
+        """
+        self._resolve_path(mds_path)
```
voegtlel, line 120: I would expect `traverse` not to modify `self`, but it does modify it when calling `_resolve_path`?

src/megatron/energon/metadataset/metadataset_v2.py
```diff
@@ -212,42 +294,55 @@ def post_initialize(self, mds_path: Optional[EPath] = None) -> None:
...
+            split_part: Split inherited from the parent traversal. If this reference defines its
+                own split override, that split takes precedence for nested traversal and the
+                returned leaf reference.
+
+        Returns:
+            A single leaf `TraversedDatasetReference` for direct dataset references, or the
+            flattened traversal result of the nested metadataset when this reference points to one.
+        """
+
+        self._resolve_path(mds_path)
```
voegtlel, line 350: Same here, there is modifications of `self` happening here.

src/megatron/energon/epathlib/epath.py
philipp-fischer: Please look at the `def _internal_str_path` property

voegtlel (changes requested):

src/megatron/energon/epathlib/epath.py
```diff
@@ -189,14 +236,26 @@ def parent(self) -> "EPath":
...
             return self._internal_str_path
         int_path_str = str(self.internal_path)
+        if self.profile == "dss":
+            if int_path_str.startswith("/"):
+                int_path_str = int_path_str[1:]
+            return f"dss://{int_path_str}"
         return f"msc://{self.profile}{int_path_str}"
 
-    def is_local(self) -> bool:
+    def is_pure_local(self) -> bool:
```
voegtlel: I think that can be embedded in the `url` method. It's the only usage? I think it may be confusing on the outside.

src/megatron/energon/metadataset/loader_interface.py
philipp-fischer: Fixed

src/megatron/energon/metadataset/metadataset.py
philipp-fischer: Good catch. I'll fix it.

philipp-fischer: @voegtlel let's discuss this again. I think `_resolve_path` is actually harmless and implementing another path without side-effects adds a lot of code... Also `traverse()` is only called from `traverse_metadataset` which constructs this structure only for this traversal and then discards it again.

voegtlel (approved):

voegtlel (approved):

PR #210, opened 2026-03-12
Add cpu-only extra for torch cpu dependency

voegtlel:
If installing with extra [cpu-only], the torch dependency will switch to using the cpu-only version to reduce dependency size.


pyproject.toml
```diff
@@ -111,3 +117,12 @@ exclude = [
   "docs",
 ]
 
+[tool.uv.sources]
+torch = [
```
philipp-fischer: Did you check that it doesn't also override torch to CPU when not using the "cpu-only" extra?

philipp-fischer (changes requested):

pyproject.toml
voegtlel: Okay, found out that this is actually broken. It would now always use the cpu version :-/ Reverted that change now, added instructions for `uv tool`.

philipp-fischer (approved):

PR #214, opened 2026-03-17
Use FFMPEG Index for Seeking

Queuecumber:
This PR switches from manual container parsing to using the internal FFMPEG frame index wherever possible, deprecating and removing the Fastseek interface (RIP, you served us well)

PyAV PR: https://github.com/PyAV-Org/PyAV/pull/2136 added the necessary plumbing to expose this index to the python side

Conceptually this is the same algorithm that Fastseek was using but now the index is built by ffmpeg which is much faster (and happens during container loading regardless of if we use it or not). 

The exact shape of this index depends on the container format so right now we support mp4/mov/mkv/webm with the index (as before) and there is a functionally identical probe fallback for other containers.

This should be a drop-in replacement and all tests pass but I recommend trying it on non-critical runs before bumping any production versions.


Queuecumber:
# Questions and Comments

1. In [AVReaderByFrames](https://github.com/NVIDIA/Megatron-Energon/pull/214/changes#diff-3af2c9ccb32cffd1f60a9359f6f21f38e9a99051d9a07250f11d96defdf53b7cR164) I included bounds checking that returns `None` if an out-of-bounds frame is requested. This matches the `ByPts` code path which naturally returns `None` if it can't find a frame in-bounds. Is that intentional or should this be a warning/error? It came up during the audio sync test.
1. I modified the [strided_resized](https://github.com/NVIDIA/Megatron-Energon/pull/214/changes#diff-edd6f40d14038682de2b4afd33e07e7488dcc01f982389392a59c34863ed3924R331) functional test to load the actual baseline video for comparison rather than comparing to the externally loaded mp4. This makes sure the seek is at least self-consistent rather than self-consistent and pixel-close with the mp4 decoding. This worked for mkv because the codecs were similar but was failing for avi which uses a different codec.
1. MP4/MOV are back to using the frame index, in my testing I found no issue decoding long videos like we had with Fastseek (because ffmpeg builds the index for us there's no perf hit), but we should try some out if you have any pathological cases

voegtlel (approved):

PR #216, opened 2026-03-19
Implement fixing permissions of written local files during prepare / media-prepare

voegtlel:
During:
energon prepare:
* Set `.nv-meta` folder permissions to the same as the parent folder
* Fix all `.nv-meta` files permissions, set file permissions to the same as the first shard
* For `.tar.idx` files, copy permissions from `.tar` files

energon prepare .jsonl:
* For `.jsonl.idx` files, copy permissions from `.jsonl` file

energon media-prepare:
* Use permissions from parent folder for `.nv-meta` folder, and also for the sqlite index within (`& 0o666`).

### DIFF
--- a/src/megatron/energon/flavors/jsonl/ijsonl.py
+++ b/src/megatron/energon/flavors/jsonl/ijsonl.py
@@ -2,7 +2,6 @@
 # SPDX-License-Identifier: BSD-3-Clause
 
 import struct
-from pathlib import Path
 from typing import BinaryIO, Dict, Generator, Optional, Tuple, Union
 
 from megatron.energon.edataclass import edataclass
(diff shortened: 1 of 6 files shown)


src/megatron/energon/flavors/jsonl/ijsonl.py
```diff
@@ -85,6 +87,16 @@ def close(self, finalize: bool = True):
         self.ijsonl.close()
         if finalize:
             self.tmp_name.move(self.final_name)
+            if self.jsonl_path.is_local():
```
philipp-fischer, line 89: We need to ensure this goes well with upcoming DSS protocol, which is also considered local (since it resolves to a cache folder). But the cache is read only

See https://github.com/NVIDIA/Megatron-Energon/blob/feature/dss/src/megatron/energon/epathlib/epath.py#L253

In general I think we should not do `is_local()` outside epathlib and then do something on `str(my_epath)`. The chmod functionality should go into epathlib.

philipp-fischer (changes requested):

src/megatron/energon/flavors/jsonl/ijsonl.py
voegtlel: I think it's just fine like this. If we're read-only we cannot run prepare anyways ;) So this only applies to non-readonly local paths, and only to dataprep.
We had already considered `chmod` for `EPath`, but that only works for local paths, dss does not support it (we asked). So I think that local code is good.

philipp-fischer: I still think it's broken for DSS, because `str(my_path)` will return `dss://something`.
We should not operate on `str(my_path)` in general.

voegtlel: Created `EPath.local_path()` and replaced `Path(str(path))` by `path.local_path()`. Kept the rest as-is

PR #221, opened 2026-04-08
Implement energon prepare with a remote dataset (using local temp)

### LINKED ISSUE #91: Support dataprep in object store
Since we are now working with sqlite databases for dataprep, currently this doesn't work in object storage.
We could implement this by creating the sqlite in local location (SSD/RAM) and then upload it.

voegtlel:
Fixes #91 

Implements energon prepare for remote (including media metadata preparation)

### DIFF
--- a/src/megatron/energon/epathlib/epath.py
+++ b/src/megatron/energon/epathlib/epath.py
@@ -336,6 +336,9 @@ class EPath:
 
         return str(self.internal_path.relative_to(other.internal_path))
 
+    def stat(self) -> msc.types.ObjectMetadata:
+        return self.fs.info(self._internal_str_path)
+
     @property
     def display_name(self) -> str:
         if self.profile == "dss":
--- a/src/megatron/energon/epathlib/epath.py
+++ b/src/megatron/energon/epathlib/epath.py
@@ -293,6 +293,17 @@ class EPath:
         # Prefix to be removed from found paths to remap to relative paths
         root_prefix = self._internal_str_path.lstrip("/")
 
+        if self.is_local():
+            # Optimize for local file system.
+            # DSS would fetch all file attributes as well.
+            local_path = self.local_path()
+            for path, _, files in os.walk(local_path, followlinks=True):
+                rel_path = PathlibPath(path).relative_to(local_path)
+                base_path = self / str(rel_path)
+                for file in files:
+                    yield base_path / file
+            return
+
         for obj in self.fs.list_recursive(self._internal_str_path):
             rel = obj.key
             if root_prefix:
--- a/src/megatron/energon/epathlib/epath.py
+++ b/src/megatron/energon/epathlib/epath.py
@@ -266,17 +266,15 @@ class EPath:
         return f"msc://{self.profile}{int_path_str}"
 
     def is_local(self) -> bool:
-        if self.profile == "dss":
-            # For now, a DSS path is always considered local.
-            # Note that this does not mean it exists on the local filesystem.
-            return True
-        else:
-            return self.profile == DEFAULT_PROFILE_NAME
+        # It will return a posix path if the fs is local, otherwise None
+        return self.fs.get_posix_path(self._internal_str_path) is not None
 
     def local_path(self) -> PathlibPath:
-        if not self.is_local():
+        # This resolves the path if it exists, probably ok.
+        posix_path = self.fs.get_posix_path(self._internal_str_path)
+        if posix_path is None:
             raise ValueError(f"Path {self} is not local")
-        return PathlibPath(self._internal_str_path)
+        return PathlibPath(posix_path)
 
     def is_dir(self) -> bool:
         try:
@@ -290,25 +288,46 @@ class EPath:
     def mkdir(self, exist_ok: bool = True, parents: bool = False):
         pass
 
-    def glob(self, pattern) -> Generator["EPath", None, None]:
+    def walk(self) -> Generator["EPath", None, None]:
+        """Returns all files within this path (no folders)."""
+        # Prefix to be removed from found paths to remap to relative paths
+        root_prefix = self._internal_str_path.lstrip("/")
+
+        for obj in self.fs.list_recursive(self._internal_str_path):
+            rel = obj.key
+            if root_prefix:
+                if rel.startswith(root_prefix + "/"):
+                    rel = rel[len(root_prefix) + 1 :]
+                elif rel.startswith("/" + root_prefix + "/"):
+                    rel = rel[len(root_prefix) + 2 :]
+                elif rel == root_prefix or rel == "/" + root_prefix:
+                    rel = "."
+
+            path = EPath(self)
+            path.internal_path = self._resolve(self.internal_path / PurePosixPath(rel))
+            yield path
+
+    def glob(self, pattern: str) -> Generator["EPath", None, None]:
+        """Returns all files matching the pattern within this path (no folders)."""
         search_path_pattern = (self / pattern)._internal_str_path
-        # MSC S3 glob matches keys like ``bucket/key``; a leading ``/`` breaks wcmatch (pattern
+        # MSC glob matches keys like ``bucket/key``; a leading ``/`` breaks wcmatch (pattern
         # ``/b/**`` never matches ``b/parts/x``). Returned keys may repeat the bucket prefix; strip
         # it before joining with ``internal_path`` so we do not get ``/b/b/parts/...``.
-        if not self.is_local() and search_path_pattern.startswith("/"):
-            search_path_pattern = search_path_pattern.lstrip("/")
+        search_path_pattern = search_path_pattern.lstrip("/")
 
-        root_prefix = str(self.internal_path).lstrip("/")
+        # Prefix to be removed from found paths to remap to relative paths
+        root_prefix = self._internal_str_path.lstrip("/")
 
         for path in self.fs.glob(search_path_pattern):
             assert isinstance(path, str)
 
             rel = path
-            if not self.is_local() and root_prefix:
-                pfx = root_prefix + "/"
-                if rel.startswith(pfx):
-                    rel = rel[len(pfx) :]
-                elif rel == root_prefix:
+            if root_prefix:
+                if rel.startswith(root_prefix + "/"):
+                    rel = rel[len(root_prefix) + 1 :]
+                elif rel.startswith("/" + root_prefix + "/"):
+                    rel = rel[len(root_prefix) + 2 :]
+                elif rel == root_prefix or rel == "/" + root_prefix:
                     rel = "."
 
             new_path = EPath(self)
--- a/src/megatron/energon/epathlib/epath.py
+++ b/src/megatron/energon/epathlib/epath.py
@@ -292,12 +292,27 @@ class EPath:
 
     def glob(self, pattern) -> Generator["EPath", None, None]:
         search_path_pattern = (self / pattern)._internal_str_path
+        # MSC S3 glob matches keys like ``bucket/key``; a leading ``/`` breaks wcmatch (pattern
+        # ``/b/**`` never matches ``b/parts/x``). Returned keys may repeat the bucket prefix; strip
+        # it before joining with ``internal_path`` so we do not get ``/b/b/parts/...``.
+        if not self.is_local() and search_path_pattern.startswith("/"):
+            search_path_pattern = search_path_pattern.lstrip("/")
+
+        root_prefix = str(self.internal_path).lstrip("/")
 
         for path in self.fs.glob(search_path_pattern):
             assert isinstance(path, str)
 
+            rel = path
+            if not self.is_local() and root_prefix:
+                pfx = root_prefix + "/"
+                if rel.startswith(pfx):
+                    rel = rel[len(pfx) :]
+                elif rel == root_prefix:
+                    rel = "."
+
             new_path = EPath(self)
-            new_path.internal_path = self._resolve(self.internal_path / PurePosixPath(path))
+            new_path.internal_path = self._resolve(self.internal_path / PurePosixPath(rel))
 
             yield new_path
 
--- a/src/megatron/energon/flavors/webdataset/prepare.py
+++ b/src/megatron/energon/flavors/webdataset/prepare.py
@@ -6,8 +6,10 @@ import json
 import logging
 import random
 import re
+import shutil
 import sys
 import tarfile
+import tempfile
 import uuid
 from pathlib import Path
 from typing import (
@@ -115,6 +117,7 @@ class SqliteIndexWriterAggregator(
     reset_tables: bool
     media_metadata_written: int
     progress_on_media: bool
+    sqlite_local_build_path: Optional[Path]
 
     def __init__(
         self,
@@ -127,6 +130,7 @@ class SqliteIndexWriterAggregator(
         media_filter: Optional[MediaFilterConfig] = None,
         reset_tables: bool = True,
         progress_on_media: bool = False,
+        sqlite_local_build_path: Optional[Path] = None,
     ):
         self.sqlite_path = sqlite_path
         self.total_tasks = total_tasks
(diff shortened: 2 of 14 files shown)


radulescupetru:
I see `ValueError: Path msc://default/path_to_s3_dataset is not local` error

```
Preparing filesystem dataset and computing media metadata...
  File "/home/user/miniconda3/envs/conda_env/bin/energon", line 7, in <module>
    sys.exit(main())
             ^^^^^^
  File "/home/user/miniconda3/envs/conda_env/lib/python3.12/site-packages/click/core.py", line 1485, in __call__
    return self.main(*args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/user/miniconda3/envs/conda_env/lib/python3.12/site-packages/click/core.py", line 1406, in main
    rv = self.invoke(ctx)
         ^^^^^^^^^^^^^^^^
  File "/home/user/miniconda3/envs/conda_env/lib/python3.12/site-packages/click/core.py", line 1873, in invoke
    return _process_result(sub_ctx.command.invoke(sub_ctx))
                           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/user/miniconda3/envs/conda_env/lib/python3.12/site-packages/click/core.py", line 1269, in invoke
    return ctx.invoke(self.callback, **ctx.params)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/user/miniconda3/envs/conda_env/lib/python3.12/site-packages/click/core.py", line 824, in invoke
    return callback(*args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/user/miniconda3/envs/conda_env/lib/python3.12/site-packages/megatron/energon/tools/prepare_media.py", line 110, in command
    stored = prepare_filesystem_dataset(
             ^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/user/miniconda3/envs/conda_env/lib/python3.12/site-packages/megatron/energon/media/filesystem_prepare.py", line 44, in prepare_filesystem_dataset
    root = root_path.local_path()
           ^^^^^^^^^^^^^^^^^^^^^^
  File "/home/user/miniconda3/envs/conda_env/lib/python3.12/site-packages/megatron/energon/epathlib/epath.py", line 258, in local_path
    raise ValueError(f"Path {self} is not local")
```
There's also a TODO in that method:

>  # Only supporting local file system, because sqlite does not support remote file systems.
>  # TODO: Implement remote file systems. Maybe create locally in tmp then upload?

voegtlel:
@radulescupetru sorry for taking a while, had other priorities. Now also implemented that for filesystem:// links. Can you try again?

src/megatron/energon/local_copy.py
philipp-fischer, line 49: Are we sure, that for any underlying filesystem, this refreshes the mtime of `tmp_path` to what the original remote mtime of `path` is? If not, we might be in trouble. E.g. if it always sets the current date, and someone later uploads a newer version of the file (from changed 7 days ago to changed 2 days ago).

voegtlel: I don't see the issue.

Upload 7 days ago
Sync to local 5 days ago (timestamped 5 days ago)
Update 2 days ago
Check now should reveal that the file loaded 5 days ago is older than the one 2 days ago -> redownload

Which case do you mean, that could be problematic?

philipp-fischer: - Upload file with modification time 7 days ago
- Sync to local today (timestamp of `tmp_path` will be 7 days ago or today?)
- Upload file with modification time 2 days ago
- Sync again: Might not sync if `tmp_path` already has today's date and uploaded file is older (2 days ago)

voegtlel: Okay, your're talking about uploading a file with an older timestamp >now<, which is actually possible. I added utime to set the timestamp to the one of online. This still assumes that online timestamps do not go to the past when updating a file.

src/megatron/energon/epathlib/epath.py
```diff
@@ -292,12 +292,27 @@ def mkdir(self, exist_ok: bool = True, parents: bool = False):
 
     def glob(self, pattern) -> Generator["EPath", None, None]:
         search_path_pattern = (self / pattern)._internal_str_path
+        # MSC S3 glob matches keys like ``bucket/key``; a leading ``/`` breaks wcmatch (pattern
```
philipp-fischer: Can we make sure, that this fix fixes it not just for S3-backed store but for all MSC-supported backends?
Also I think there's a bug for DSS datasets here. Note that DSS datasets are considered local but still carry the DSS protocol.

src/megatron/energon/epathlib/epath.py
```diff
@@ -292,12 +292,27 @@ def mkdir(self, exist_ok: bool = True, parents: bool = False):
...
+            if not self.is_local() and root_prefix:
+                pfx = root_prefix + "/"
+                if rel.startswith(pfx):
+                    rel = rel[len(pfx) :]
+                elif rel == root_prefix:
+                    rel = "."
+
             new_path = EPath(self)
-            new_path.internal_path = self._resolve(self.internal_path / PurePosixPath(path))
+            new_path.internal_path = self._resolve(self.internal_path / PurePosixPath(rel))
```
philipp-fischer, line 345: For DSS, the `self.internal_path` will not be the local cache path.

src/megatron/energon/media/filesystem_prepare.py
```diff
@@ -64,84 +70,93 @@ def agg_progress_fn(iterator: Iterator[int], total: int) -> Iterator[int]:
...
 
-    for dirpath, dirnames, filenames in os.walk(root, followlinks=False):
-        current_dir = Path(dirpath)
-
-        if current_dir.name == MAIN_FOLDER_NAME:
-            dirnames[:] = []
-            continue
+    if root.is_local():
+        paths = (
+            EPath(f"{path}/{file}")
```
philipp-fischer: Composing a new local path like this, will break/kill the DSS reference.
Probably, we don't want to support DSS anyway since for now it's read-only. Still, I think recomposing like this is not good. It should be like `path / file` not `EPath(f"{path}/{file}")`. And shouldn't we provide an EPath-based walk method instead of using os.walk outside?

src/megatron/energon/media/filesystem_prepare.py
```diff
@@ -151,18 +166,17 @@ def _collect_media_files(
...
     media_filter: MediaFilterConfig,
 ) -> Iterator[IndexMediaMetadata]:
-    file_path = Path(file_path)
     extracted_metadata = media_filter.extract_metadata(file_path)
     if extracted_metadata is None:
         return
 
     stored_type, metadata_json = serialize_media_metadata(extracted_metadata)
-    entry_key = file_path.relative_to(root).as_posix()
+    entry_key = file_path.relative_to(root)
```
philipp-fischer, line 184: I think this will fail if one is newly composed local path and the other is a DSS root.

src/megatron/energon/media/filesystem_prepare.py
```diff
@@ -64,84 +70,93 @@ def agg_progress_fn(iterator: Iterator[int], total: int) -> Iterator[int]:
...
-
-            if not consider_all and not media_filter.should_consider_media(filename):
-                continue
+        if not consider_all and not media_filter.should_consider_media(file.name):
+            continue
 
-            candidate = current_dir / filename
-            if MAIN_FOLDER_NAME in candidate.parts:
-                continue
+        if ("/" + MAIN_FOLDER_NAME + "/") in file.url:
```
philipp-fischer: It kind of relies on slashes as path seprator here

src/megatron/energon/media/filesystem_prepare.py
```diff
@@ -43,18 +46,21 @@ def prepare_filesystem_dataset(
         Number of metadata entries written to the database.
     """
 
-    # Only supporting local file system, because sqlite does not support remote file systems.
-    # TODO: Implement remote file systems. Maybe create locally in tmp then upload?
-    root = root_path.local_path()
-    assert root.is_dir(), f"Expected directory for filesystem dataset, got {root}"
-    assert root.is_absolute(), f"Filesystem dataset path must be absolute: {root}"
+    assert not root_path.is_file(), f"Expected directory for filesystem dataset, got {root_path}"
```
philipp-fischer, line 51: For local non-existent dirs this would not trigger. Before, we explicitly checked for an existing folder.

src/megatron/energon/media/filesystem_prepare.py
```diff
@@ -64,84 +70,93 @@ def agg_progress_fn(iterator: Iterator[int], total: int) -> Iterator[int]:
...
 
         progress_bar = tqdm(total=None, unit="file", desc="Collecting media files")
 
-    for dirpath, dirnames, filenames in os.walk(root, followlinks=False):
-        current_dir = Path(dirpath)
-
-        if current_dir.name == MAIN_FOLDER_NAME:
-            dirnames[:] = []
-            continue
+    if root.is_local():
```
philipp-fischer: In the old code, we exclduded `.nv-meta` during traversal. Here we are collecting those files again but dismissing them later. If there are not to many, I think it should be fine.

src/megatron/energon/tools/prepare_media.py
```diff
@@ -44,13 +47,23 @@
...
+    ),
+)
 def command(
     path: EPath,
     progress: bool,
     num_workers: int,
     media_metadata_by_glob: str | None,
     media_metadata_by_header: bool,
     media_metadata_by_extension: bool,
+    tmp_path: Optional[Path],
```
philipp-fischer, line 66: This is passed to WebDataset prep but not for filesystem prep below (`prepare_filesystem_dataset`)

src/megatron/energon/flavors/webdataset/prepare.py
```diff
@@ -744,6 +782,7 @@ def add_media_metadata(
         media_filter: MediaFilterConfig,
         workers: int = 32,
         progress_fn: Callable[[Iterator[Any], int], Iterator[T]] = (lambda x, y: x),
+        index_sqlite_tmp_path: Optional[Path] = None,
```
philipp-fischer, line 785: `index_sqlite_tmp_path` isn't clearly documented/defined. Int he CLI it says it should be a folder, but below it expects a path to an sqlite file.

src/megatron/energon/flavors/webdataset/prepare.py
```diff
@@ -507,6 +521,9 @@ def prepare_dataset(
             tar_index_only: Only create tar-index, then exit
             media_filter: Media filter configuration
             fix_duplicates: If True, fix duplicate keys in the dataset by renaming the files in the shards.
+            index_sqlite_tmp_path: When ``parent_path`` is remote, directory used to build ``index.sqlite``
```
philipp-fischer: Again, here it says directory, but below it expects a file path. (passed to `SqliteIndexWriterAggregator`)

src/megatron/energon/local_copy.py
```diff
@@ -33,15 +35,19 @@ def ensure_local_copy(path: EPath) -> EPath:
     final_path = LOCAL_COPY_TMP_DIR / f"{digest}.bin"
     lock_path = final_path.with_suffix(".lock")
     tmp_path = final_path.with_suffix(".part")
+    mod_time = path.stat().last_modified.timestamp()
 
     # Block until lock is free
     with FileLock(lock_path, timeout=60 * 5):
-        if final_path.exists():  # someone else already produced it
+        # someone else already produced it
+        if final_path.exists() and final_path.stat().st_mtime >= mod_time:
```
philipp-fischer, line 43: Wondering if second-level granularity is sufficient. Sub-second updates would be lost.
Also one more concern: The S3 emulator used for testing, doesn't handle timestamps correctly so any issues with the file stats cannot be covered in tests.

philipp-fischer (changes requested):

src/megatron/energon/epathlib/epath.py
voegtlel: Replaced this universally

src/megatron/energon/epathlib/epath.py
voegtlel: Sure, but I'd want to keep it `dss://`, which it would be.

src/megatron/energon/media/filesystem_prepare.py
voegtlel: Prepare doesn't work on dss, because it's readonly anyways.

src/megatron/energon/media/filesystem_prepare.py
voegtlel: Prepare doesn't work on dss, because it's readonly anyways.

Nevertheless, this is being tested and working.

src/megatron/energon/media/filesystem_prepare.py
voegtlel: Yes, I believe we actually do that in more places, also this is `.url`, i'd expect slashes here for sure.

src/megatron/energon/media/filesystem_prepare.py
voegtlel: Yes, because glob anyways cant filter on that beforehand.

src/megatron/energon/media/filesystem_prepare.py
voegtlel: Why this way: Because `root.glob` for `**/*` is much less efficient than `os.walk`. It's not even a generator, it will create the full list ahead, while this here is an actual generator.

voegtlel: Moved the walk to EPath.

src/megatron/energon/local_copy.py
voegtlel: S3 emulator now handles timestamps correctly. I believe we can disregard sub-second changes for now? Is it likely that people do sub-second updates on other than local storage (e.g. S3)?

src/megatron/energon/epathlib/epath.py
```diff
@@ -290,25 +288,46 @@ def is_file(self) -> bool:
     def mkdir(self, exist_ok: bool = True, parents: bool = False):
         pass
 
-    def glob(self, pattern) -> Generator["EPath", None, None]:
+    def walk(self) -> Generator["EPath", None, None]:
+        """Returns all files within this path (no folders)."""
+        # Prefix to be removed from found paths to remap to relative paths
+        root_prefix = self._internal_str_path.lstrip("/")
+
+        for obj in self.fs.list_recursive(self._internal_str_path):
```
philipp-fischer, line 307: Either us or MSC team needs to fix this for local paths. This is way slower than os.walk. Maybe for now we should have os.walk here explicitly for local paths. In MSC they do os.listdir + sorting + isdir/isdile plus object construction for each file, i.e. lots of overhead.
But let's make sure we still preserve DSS URLs when using os.walk.

src/megatron/energon/media/filesystem_prepare.py
philipp-fischer: Still unresolved

src/megatron/energon/flavors/webdataset/prepare.py
philipp-fischer, line 584: This will evaluate to True for DSS caches, because they are local. Can we maybe catch the read-only case somewhere early to get a useful error? (also for the other prepare entry points)

philipp-fischer (changes requested):

src/megatron/energon/media/filesystem_prepare.py
voegtlel: We cannot reliably check for `is_dir` with object store. Thus this check is gone. I added a raise if there were no files found.

src/megatron/energon/epathlib/epath.py
voegtlel: Okay, I optimized it for us. I actually think it doesn't hurt too much. This is a generator anyways, and processing the media files requires the metadata anyways, so no advantage optimizing.

src/megatron/energon/flavors/webdataset/prepare.py
voegtlel: As explained in the other comments: DSS is readonly anyways. Also, this should actually resolve to the unix path (see `SqliteIndexWriter`: `sqlite_path.local_path()` should resolve well for DSS), so all fine?

philipp-fischer: So this code will try to write into the DSS local cache and would then fail with permission error. Is that the behavior we want? Or rather catch it earlier?

voegtlel: Yes, would not catch earlier

PR #223, opened 2026-04-14
Fixing duplicate SkipSample class

bbuschkaemper:
The `SkipSample` class is defined two times:
- In `megatron.energon.errors`: [link](https://github.com/NVIDIA/Megatron-Energon/blob/d84075adb2d787d37ad8f9b958fde5445c002188/src/megatron/energon/errors.py#L13)
- In`megatron.energon.wrappers`: [link](https://github.com/NVIDIA/Megatron-Energon/blob/d84075adb2d787d37ad8f9b958fde5445c002188/src/megatron/energon/wrappers/skip.py#L5)

Both definitions are distinct error classes, although named the same. However, only the `megatron.energon.errors` definition is caught by megatron-energon [here](https://github.com/NVIDIA/Megatron-Energon/blob/d84075adb2d787d37ad8f9b958fde5445c002188/src/megatron/energon/errors.py#L115) and [here](https://github.com/NVIDIA/Megatron-Energon/blob/d84075adb2d787d37ad8f9b958fde5445c002188/src/megatron/energon/errors.py#L161).

The main `megatron.energon` init file is also importing the invalid `megatron.energon.wrappers` definition of the `SkipSample` class, leading to crashes in the run log when importing from the main package definition.

This PR deprecates the invalid duplicate and points the `megatron.energon` init file to the correct subpackage definition of the `SkipSample` class.

### DIFF
--- a/src/megatron/energon/__init__.py
+++ b/src/megatron/energon/__init__.py
@@ -24,7 +24,7 @@ from megatron.energon.cache import (
     SystemFileStore,
 )
 from megatron.energon.edataclass import edataclass
-from megatron.energon.errors import log_exception, reraise_exception
+from megatron.energon.errors import SkipSample, log_exception, reraise_exception
 from megatron.energon.flavors import (
     BaseCoreDatasetFactory,
     BaseWebdatasetFactory,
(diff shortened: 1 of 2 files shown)


philipp-fischer:
Thanks Björn for your contribution 🙏🏻
It's good to have this duplicate fixed. I guess at some point we will remove it entirely.

PR #224, opened 2026-04-15
Removed unneeded sqlite index.

bbuschkaemper:
SQLite implicitly creates an index due to `sample_key` being unique. The added index is an unnecessary duplicate resulting in a slowdown, especially for very large (i.e. >1B row) datasets.


philipp-fischer:
@lvoegtle What do you think, do we really need a unit test for sqlite indices?

bbuschkaemper:
@philipp-fischer I could remove the unit test if you want, was prob an overkill

voegtlel:
> @lvoegtle What do you think, do we really need a unit test for sqlite indices?

Yes, I think we can omit the new additional unittest. This is tested quite thoroughly through the e2e tests.

src/megatron/energon/flavors/webdataset/indexing.py
```diff
@@ -219,10 +219,7 @@ def close(self):
 
         if self.enable_sample_tables:
             # Create the index after adding all the samples for better speed
-            # Index on sample_key for fast lookups
-            self.db.execute(
-                "CREATE UNIQUE INDEX IF NOT EXISTS idx_samples_sample_key ON samples(sample_key)"
```
voegtlel, line 242: Imho, this is used by `SqliteIndexReader.get_sample_pointer_by_key`?

voegtlel (changes requested):
I believe this is actually needed

src/megatron/energon/flavors/webdataset/indexing.py
bbuschkaemper: I don't think the explicit `idx_samples_sample_key` index is needed here. `sample_key` is already declared as `TEXT NOT NULL UNIQUE`, and in SQLite that `UNIQUE` constraint is normally backed by an implicit unique index (see [link](https://www.sqlite.org/lang_createtable.html#uniqueconst)). So `SqliteIndexReader.get_sample_pointer_by_key` should still have an indexed lookup for `WHERE sample_key = ?` even after removing the explicit index.

In other words, the removed index is a duplicate of the implicit one, so it adds write/storage overhead without changing correctness. I kept the old `DROP INDEX IF EXISTS ...` for backwards cleanup of existing DBs.

If the schema-level test feels too implementation-specific, I’m happy to drop that one and keep only the behavioral coverage that lookup-by-key still works and duplicate keys still raise.

voegtlel: Ah sorry, you're absolutely right. Fine by me, but would prefer to omit the additional tests.

bbuschkaemper: Done

voegtlel (approved):

voegtlel:
Thank you for this fix! Merging when Tests succeed.

PR #225, opened 2026-04-23
Batch WebDataset index writes during prepare_dataset

bbuschkaemper:
## Summary

Reduce `prepare_dataset()` indexing overhead by batching worker output and SQLite writes, giving massive speed increase on large, multi-worker scenarios.

## Motivation

`prepare_dataset()` already parallelizes shard scanning, but workers were still sending one item at a time to a single SQLite writer process. That left a lot of overhead in:

- multiprocessing queue traffic
- Python object handling in the aggregator
- one-row-at-a-time SQLite inserts

For larger datasets, that becomes a noticeable bottleneck during dataset preparation (prepare took ~30h for a 2B sequence dataset with 128 workers on a 2TB RAM + 128-core VM).

## Changes

- add an internal `IndexBatch` payload for batched worker (aggregator communication)
- batch sample, sample-part, and media-metadata rows in the worker before flushing
- switch the SQLite writer to bulk inserts via `executemany()`
- preserve duplicate-key reporting for batched sample inserts
- unit test for batched duplicate reporting

 ## Benchmark

Synthetic benchmark:
- 400k samples
- 40 shards

| workers | before | after |
| --- | ---: | ---: |
| 1 | 67.247s | 40.137 |
| 4 | 24.707s | 11.294s |
| 8 | 19.593s | 6.451s |

### DIFF
--- a/src/megatron/energon/flavors/webdataset/indexing.py
+++ b/src/megatron/energon/flavors/webdataset/indexing.py
@@ -137,38 +137,6 @@ class SqliteIndexWriter:
                 """
             )
 
-    def append_sample(
-        self,
-        tar_file_id: int8,
-        sample_key: str,
-        sample_index: int,
-        byte_offset: Optional[int],
-        byte_size: Optional[int],
-    ):
-        """
-        Adds a new sample row to the samples table.
-
-        Args:
-            tar_file_id: The index of the tar file in the reader.
-            sample_key: The key of the sample.
-            sample_index: The index of the sample in the tar file.
-            byte_offset: The byte offset of the sample in the tar file.
-            byte_size: The size of the sample in the tar file.
-        """
-
-        assert self.db is not None, "Database is closed"
-
-        try:
-            self.db.execute(
-                """
-                INSERT INTO samples (tar_file_id, sample_key, sample_index, byte_offset, byte_size)
-                VALUES (?, ?, ?, ?, ?)
-                """,
-                (tar_file_id, sample_key, sample_index, byte_offset, byte_size),
-            )
-        except sqlite3.IntegrityError as exc:
-            raise DuplicateSampleKeyError(sample_key) from exc
-
     def append_samples(
         self,
         rows: Sequence["IndexSample"],
@@ -207,26 +175,6 @@ class SqliteIndexWriter:
             self.db.execute(f"RELEASE SAVEPOINT {savepoint_name}")
             raise DuplicateSampleKeyError(self._find_duplicate_sample_key(rows)) from exc
 
-    def append_part(
-        self,
-        tar_file_id: int8,
-        sample_index: int,
-        part_name: str,
-        content_byte_offset: int,
-        content_byte_size: int,
-    ):
-        """Adds a new part row to the samples table."""
-
-        assert self.db is not None, "Database is closed"
-
-        self.db.execute(
-            """
-            INSERT INTO sample_parts (tar_file_id, sample_index, part_name, content_byte_offset, content_byte_size)
-            VALUES (?, ?, ?, ?, ?)
-            """,
-            (tar_file_id, sample_index, part_name, content_byte_offset, content_byte_size),
-        )
-
     def append_parts(
         self,
         rows: Sequence["IndexSamplePart"],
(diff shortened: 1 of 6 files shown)


src/megatron/energon/flavors/webdataset/prepare.py
```diff
@@ -101,6 +101,15 @@ class IndexShardInfo(IndexAggregatable):
     parts: Set[str]
 
 
+@edataclass
+class IndexBatch(IndexAggregatable):
+    """A chunk of rows produced by one worker and consumed by the SQLite aggregator."""
+
+    samples: Tuple[Tuple[int, str, int, int, int], ...]
```
voegtlel: Can we reuse the `IndexSample/IndexSamplePart/IndexMediaMetadata` here? Performance should be quite close to tuples, and it gives clear typing.

src/megatron/energon/flavors/webdataset/prepare.py
```diff
@@ -163,11 +172,38 @@ def on_item(
...
+            if item.sample_parts:
+                self.writer.append_parts(item.sample_parts)
+            if item.media_metadata:
+                self.writer.append_media_metadata_batch(item.media_metadata)
+            if item.samples or item.sample_parts or item.media_metadata:
+                self.had_update = True
+            self.media_metadata_written += len(item.media_metadata)
+            if self.progress_on_media:
+                self._advance_progress(count=len(item.media_metadata))
+        elif isinstance(item, IndexSample):
```
voegtlel: I believe this and also the following objects are now obsolete, and can thus be removed

voegtlel (changes requested):
Hi @bbuschkaemper , thank you very much for this finding and the suggested improvements! I like the solution to batch.

One global suggestion for a simpler implementation:
Instead of implementing this in the processing level (i.e. each aggregator), we could implement it in the `AggregatorPool`. I.e. in the `_worker`, we could batch the messages e.g. according to a `batch_size` (to be set in `__init__`), and in `_aggregator_run`, we pass that batch to `self.aggregator.on_item(item_batch, self)`. That would allow keeping the data models as before, but would of course require that the grouping&batching to happen in the `SqliteIndexWriterAggregator.on_item`. I believe this would be slightly better capsulated, would you agree?

bbuschkaemper:
@voegtlel Good idea, implemented that. Can you recheck? Especially prepare.py should be clean now.

src/megatron/energon/flavors/webdataset/indexing.py
```diff
@@ -159,9 +166,47 @@ def append_sample(
...
+                ),
+            )
+            self.db.execute(f"RELEASE SAVEPOINT {savepoint_name}")
+        except sqlite3.IntegrityError as exc:
+            # executemany() may already have inserted earlier rows in the batch
+            self.db.execute(f"ROLLBACK TO SAVEPOINT {savepoint_name}")
+            self.db.execute(f"RELEASE SAVEPOINT {savepoint_name}")
+            raise DuplicateSampleKeyError(self._find_duplicate_sample_key(rows)) from exc
+
     def append_part(
```
voegtlel: Voting for removing the `append_sample`, `append_part` and `append_media_metadata`. They are obsolete now.

voegtlel (changes requested):
Looks good to me now. Just requesting this one cleanup of the old methods :)

src/megatron/energon/flavors/webdataset/indexing.py
bbuschkaemper: @voegtlel Removed them in latest commit. Should be cautious since this is technically a breaking change, no?

voegtlel: We consider these classes internal, and they are not exposed in the main API. So this should be fine.

voegtlel (approved):

bbuschkaemper:
@voegtlel Could you rerun workflows? ruff failed on unused import I missed, just fixed it.

philipp-fischer (changes requested):
Please also adapt `filesystem_prepare.py` to use the new batching (with a batch size > 1) and test that it works. Leaving it with the current default batch_size=1 would make it slower than it was before.
So the currently private `_INDEX_BATCH_SIZE` should become a shared setting across those modules (maybe in `config.py`). Alternatively, we could change the default `batch_size: int = 1,` to this value, since we would always want batching by default.

bbuschkaemper:
> Please also adapt `filesystem_prepare.py` to use the new batching (with a batch size > 1) and test that it works. Leaving it with the current default batch_size=1 would make it slower than it was before. So the currently private `_INDEX_BATCH_SIZE` should become a shared setting across those modules (maybe in `config.py`). Alternatively, we could change the default `batch_size: int = 1,` to this value, since we would always want batching by default.

Done. I added `INDEX_BATCH_SIZE` to `config.py` and use it in webdataset_prepare and filesystem_prepare

philipp-fischer (approved):
Thanks

PR #226, opened 2026-04-28
Adjust DSS cache path to not contain `@` but `/` as a version separator

philipp-fischer:


src/megatron/energon/epathlib/epath.py
```diff
@@ -41,14 +41,23 @@ class EPath:
...
+        "dss_dataset_version",
+    )
 
     # The path without the protocol/profile. Can also be in S3 for example
     internal_path: PurePosixPath
     # The profile used to access the file system
     profile: str
     # The file system
     fs: msc.StorageClient
+    # DSS dataset name and version, extracted from dss://name@version URLs.
```
voegtlel: Why don't we change it to dss://name/version? Then we could simply omit all these changes here, which add more special handling just for supporting dss.

src/megatron/energon/epathlib/epath.py
```diff
@@ -41,14 +41,23 @@ class EPath:
...
+    dss_dataset_name: Optional[str]
+    dss_dataset_version: Optional[str]
```
voegtlel: Why do we need these extra fields? Can't we just modify the printing? It seems to add a lot of weight, especially for people not using dss

voegtlel (changes requested):

voegtlel (approved):

PR #230, opened 2026-05-19
Expose enable_sample_tables in prepare_dataset and as --no-sample-tables CLI flag

### LINKED ISSUE #231: Add option to skip SQLite samples/sample_parts tables during prepare (multi-hour cost on 100M+ sample datasets)
## Problem

On very large WebDataset materialisations, the SQLite indexing phase of `energon prepare` dominates total runtime. On a 244M-row / 4000-shard text-only dataset, the indexing step has been the bottleneck — we hit a 13-hour K8s pod timeout in the SQLite phase (the tar-write phase finished in ~8h).

Profile-by-inspection: with `~488M` rows across `samples` + `sample_parts` and `UNIQUE(sample_key)` enforcement, the dominant costs are (1) single-writer `INSERT` throughput, (2) UNIQUE-constraint checks on a multi-GB btree, and (3) the post-load `CREATE INDEX` over `samples(sample_key)` and the two `sample_parts` indexes.

For datasets that are consumed purely sequentially via `ShardInfosITarReader` (e.g. large-scale text pretraining), the `samples` / `sample_parts` tables are not queried at training time — checkpoint/resume uses integer `SliceState` offsets resolved via `.info.json` + `.tar.idx`. So the multi-hour SQLite build is producing tables that are never read.

`SqliteIndexWriter` already accepts `enable_sample_tables=False`, but that knob is not plumbed through `BaseWebdatasetFactory.prepare_dataset()` or the `energon prepare` CLI, so end-users can't opt out.

## What the existing `--tar-index-only` flag is *not*

`--tar-index-only` does **not** help here. Reading `prepare_dataset` end-to-end: the `SqliteIndexWriterAggregator` is constructed unconditionally, the worker pool runs and feeds all `IndexSample` / `IndexSamplePart` items into the SQLite `INSERT`s, and only *after* that does the function check `tar_index_only` and return early — skipping just the trivial `split.yaml` / `dataset.yaml` write. The CLI further requires `.info.json` to already exist (it reads `get_dataset_info(path)` first), so `--tar-index-only` is for *re-indexing* an already-prepared dataset, not for skipping indexing.

## Proposal

Expose `enable_sample_tables` through `BaseWebdatasetFactory.prepare_dataset()` (default `True` — backwards compatible) and add a corresponding `--no-sample-tables` CLI flag. When `False`:

- `samples` / `sample_parts` tables are not created.
- `INSERT`s become no-ops.
- The post-load `CREATE INDEX` calls are skipped (already gated in `close()`).
- `.tar.idx`, `.info.json`, and split config are still produced — the integer-indexed loader continues to work.

Known limitations of `--no-sample-tables` (which the PR documents):

- Polylithic dataset joins (`join_dataset_loader.py` SQL `JOIN` on `samples.sample_key`) — will fail.
- `WebdatasetFileStore` / `SqliteITarEntryReader` — used by `as_file_store()`, `energon mount`, and aux-data access for `CrudeSample` datasets — will fail at first key lookup.
- Any user code that queries the SQLite sample tables.

All failures are loud (`OperationalError: no such table: samples`), not silent.

## PR

Implemented in #230.

pei-li-hedgehog:
Closes #231.

## Summary

`SqliteIndexWriter` already supports `enable_sample_tables=False` to skip creating the `samples` and `sample_parts` tables, but that knob isn't plumbed through `BaseWebdatasetFactory.prepare_dataset()` or the `energon prepare` CLI. On very large datasets (100M+ samples) the SQLite inserts and the post-insert btree builds dominate preparation time. For datasets consumed purely by the integer-indexed loader (`ShardInfosITarReader`), the `samples` table isn't queried at training time — checkpoint/resume uses `SliceState` integer offsets resolved via `.info.json` + `.tar.idx`.

Concretely, on a 244M-row / 4000-shard text webdataset we hit a 13-hour K8s pod timeout in the indexing phase. We need a way to skip the SQLite cost for monolithic pretraining datasets while keeping everything the integer-indexed loader needs (`.tar.idx`, `.info.json`, split config).

## Changes

- `BaseWebdatasetFactory.prepare_dataset()` accepts `enable_sample_tables=True` (default — backwards compatible), passes through to `SqliteIndexWriterAggregator`.
- `energon prepare --no-sample-tables` exposes the same option. Mutually exclusive with `--tar-index-only` (which operates on an already-prepared dataset, different semantics).
- `append_samples` / `append_parts` become no-ops when the writer was constructed with `enable_sample_tables=False`, so producers can keep emitting rows unmodified.
- `SqliteIndexWriter` always drops stale `samples` / `sample_parts` when `reset_tables=True`, regardless of whether the new run intends to repopulate them. Avoids stale tables when re-preparing with `--no-sample-tables` after a previous full prepare.

## Out of scope

Sample-key lookups (`WebdatasetFileStore` / `SqliteITarEntryReader`), polylithic joins (which `INNER JOIN ... ON samples.sample_key = ...` at metadataset-prepare time), and aux-data access on `CrudeSample` datasets all rely on the SQLite sample tables and will not work for datasets prepared with `--no-sample-tables`. That's the intended trade-off — documented in the new docstring/help text. Failures are loud (clear SQLite `no such table` errors), not silent.

A larger follow-up would also skip emitting `IndexSample` / `IndexSamplePart` items from `_preprocess_tar` (saving IPC overhead in addition to SQLite cost). Out of scope here.

## Testing

All four existing `test_prepare_dataset*` tests pass (backwards compatibility).

Two new tests:

- `test_prepare_dataset_no_sample_tables` — covers the prepare path:
  - `--no-sample-tables` prepare succeeds; `.info.json` and per-tar `.tar.idx` are produced.
  - `index.sqlite` exists but does **not** contain the `samples` / `sample_parts` tables.
  - `--no-sample-tables` + `--tar-index-only` is rejected with the documented `UsageError`.
  - Constructing a `SqliteIndexReader` against the prepared dataset and calling sample-key methods (`get_sample_pointer_by_key`, `get_sample_count`, `get_sample_part`) raises `MissingSamplesTableError` with a descriptive message — not the raw `sqlite3.OperationalError: no such table: samples`.

- `test_prepare_dataset_no_sample_tables_save_restore` — covers the training-time checkpoint/resume path on a `--no-sample-tables` dataset:
  - Reference: uninterrupted iteration of 20 samples.
  - Save state mid-stream after 10 samples via `save_state_rank()`; continue iterating to capture the next 10 (`post_save`).
  - Build a fresh loader, `restore_state_rank(state)`, iterate 10 samples (`post_restore`).
  - Asserts `first_half + post_save == reference` and `post_restore == post_save`.
  - This empirically verifies that `ShardInfosITarReader` + `SliceState` (the integer-indexed loader/state path) work end-to-end on a dataset prepared without the SQLite samples tables.

`ruff check` and `ruff format --check` clean on all touched files.

### DIFF
--- a/docs/source/advanced/data_prep_api.md
+++ b/docs/source/advanced/data_prep_api.md
@@ -71,22 +71,4 @@ if __name__ == "__main__":
 
 ## Skipping the SQLite samples tables for very large datasets
 
-`prepare_dataset` accepts an `enable_sample_tables` kwarg (default `True`). On very large datasets (100M+ samples) the SQLite inserts and post-load btree builds for the `samples` and `sample_parts` tables can dominate preparation runtime. If the dataset is consumed purely sequentially via the integer-indexed loader (`ShardInfosITarReader`), those tables are never queried at training time, and you can skip populating them by passing `enable_sample_tables=False`:
-
-```python
-BaseWebdatasetFactory.prepare_dataset(
-    path,
-    all_tars,
-    split_parts_ratio=split_parts_ratio,
-    progress_fn=progress_fn,
-    workers=num_workers,
-    enable_sample_tables=False,
-)
-```
-
-`.tar.idx`, `.info.json` and the split config are still produced.
-
-```{admonition} Trade-off
-:class: warning
-With `enable_sample_tables=False`, sample-key lookups are unavailable. This breaks polylithic dataset joins (built via SQL `JOIN` over the `samples` tables), `as_file_store()` / `WebdatasetFileStore` access (used for aux-data on crude datasets and by `energon mount`), and any direct `SqliteIndexReader` queries. Failures are loud (`sqlite3.OperationalError: no such table: samples`).
-```
+Pass `enable_sample_tables=False` to `prepare_dataset` (default `True`) to skip populating the `samples` and `sample_parts` tables. Useful when preparation is bottlenecked by SQLite indexing on very large datasets. A dataset prepared this way cannot be used as an [auxiliary dataset](aux-data).
--- a/docs/source/advanced/data_prep_api.md
+++ b/docs/source/advanced/data_prep_api.md
@@ -68,3 +68,25 @@ Then, run the script:
 if __name__ == "__main__":
     prepare_one_dataset(Path("/path/to/dataset"), 16, Path("/path/to/template_dir"))
 ```
+
+## Skipping the SQLite samples tables for very large datasets
+
+`prepare_dataset` accepts an `enable_sample_tables` kwarg (default `True`). On very large datasets (100M+ samples) the SQLite inserts and post-load btree builds for the `samples` and `sample_parts` tables can dominate preparation runtime. If the dataset is consumed purely sequentially via the integer-indexed loader (`ShardInfosITarReader`), those tables are never queried at training time, and you can skip populating them by passing `enable_sample_tables=False`:
+
+```python
+BaseWebdatasetFactory.prepare_dataset(
+    path,
+    all_tars,
+    split_parts_ratio=split_parts_ratio,
+    progress_fn=progress_fn,
+    workers=num_workers,
+    enable_sample_tables=False,
+)
+```
+
+`.tar.idx`, `.info.json` and the split config are still produced.
+
+```{admonition} Trade-off
+:class: warning
+With `enable_sample_tables=False`, sample-key lookups are unavailable. This breaks polylithic dataset joins (built via SQL `JOIN` over the `samples` tables), `as_file_store()` / `WebdatasetFileStore` access (used for aux-data on crude datasets and by `energon mount`), and any direct `SqliteIndexReader` queries. Failures are loud (`sqlite3.OperationalError: no such table: samples`).
+```
--- a/docs/source/basic/data_prep.md
+++ b/docs/source/basic/data_prep.md
@@ -688,13 +688,13 @@ The `media_metadata` table is used to store the media metadata for the selected 
 
 #### Skipping the samples tables for very large datasets
 
-For very large datasets (100M+ samples), populating the `samples` and `sample_parts` tables can dominate `energon prepare` runtime. If the dataset will only be consumed sequentially (not as an auxiliary dataset), you can skip these tables:
+For very large datasets (100M+ samples), populating the `samples` and `sample_parts` tables can dominate `energon prepare` runtime. If the dataset will not be used as an auxiliary dataset or mounted, you can skip these tables:
 
 ```sh
 > energon prepare --no-sample-tables /path/to/dataset
 ```
 
-A dataset prepared this way cannot be used as an [auxiliary dataset](aux-data).
+A dataset prepared this way cannot be used as an [auxiliary dataset](aux-data) or with `energon mount`.
 
 (data-on-disk-jsonl)=
 ## Dataset Format on Disk for JSONL Datasets
--- a/docs/source/basic/data_prep.md
+++ b/docs/source/basic/data_prep.md
@@ -686,6 +686,28 @@ The `media_metadata` table is used to store the media metadata for the selected 
 | 00002.mp4 | ... | ... |
 
 
+#### Skipping the samples tables for very large datasets
+
+On very large datasets (100M+ samples) the `samples` and `sample_parts` table inserts, combined with the post-load btree index builds over the unique `sample_key`, can dominate `energon prepare` runtime. For datasets that are consumed purely sequentially via the integer-indexed loader (`ShardInfosITarReader` — the default loader for monolithic webdatasets), those tables are never queried at training time: shard cumulative counts come from `.info.json`, and per-sample byte offsets come from the per-tar `.tar.idx` files. Checkpoint/resume uses integer `SliceState` offsets resolved through the same path.
+
+If you do not need sample-key lookups, you can skip populating the SQLite sample tables:
+
+```sh
+> energon prepare --no-sample-tables /path/to/dataset
+```
+
+This still produces `.tar.idx`, `.info.json` and the split config — everything the integer-indexed loader needs.
+
+```{admonition} Trade-off
+:class: warning
+Datasets prepared with `--no-sample-tables` cannot be used with features that look samples up by `sample_key`. In particular:
+
+- [Polylithic datasets](aux-data) — the join is built by an SQL `JOIN` over the `samples` tables.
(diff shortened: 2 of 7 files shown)


docs/source/basic/data_prep.md
```diff
@@ -686,6 +686,28 @@ The `media_metadata` table is used to store the media metadata for the selected
 | 00002.mp4 | ... | ... |
 
 
+#### Skipping the samples tables for very large datasets
```
voegtlel, line 689: I think we should make this much shorter. The essence is that we can skip the tables above and as a consequence cannot use the dataset as aux.

docs/source/advanced/data_prep_api.md
```diff
@@ -68,3 +68,25 @@ Then, run the script:
 if __name__ == "__main__":
     prepare_one_dataset(Path("/path/to/dataset"), 16, Path("/path/to/template_dir"))
 ```
+
+## Skipping the SQLite samples tables for very large datasets
```
voegtlel, line 72: Let's shorten this, less details

docs/source/advanced/data_prep_api.md
```diff
@@ -68,3 +68,25 @@ Then, run the script:
...
+    workers=num_workers,
+    enable_sample_tables=False,
+)
+```
+
+`.tar.idx`, `.info.json` and the split config are still produced.
+
+```{admonition} Trade-off
+:class: warning
+With `enable_sample_tables=False`, sample-key lookups are unavailable. This breaks polylithic dataset joins (built via SQL `JOIN` over the `samples` tables), `as_file_store()` / `WebdatasetFileStore` access (used for aux-data on crude datasets and by `energon mount`), and any direct `SqliteIndexReader` queries. Failures are loud (`sqlite3.OperationalError: no such table: samples`).
```
voegtlel: The primary thing is that it cannot be used as aux.

voegtlel (changes requested):
Hi @pei-li-hedgehog , thank you for this feature! Basically looks good to me, I'd just reduce the docs a bit, to be more on the spot.

voegtlel:
One more thing: We may need to additionally add a proper error message when loading such a dataset as aux fails, i.e. if the sqlite doesn't contain the `samples` table, but we try to load it, explaining that the dataset was created without an index, and that it needs to be prepared accordingly.

pei-li-hedgehog:
Thanks for the review @voegtlel! All four points addressed. 

Good that you mentioned about `adding a proper error message when loading such a dataset as aux fails`.  I've made the following change:

- New MissingSamplesTableError in indexing.py. Message names the cause (--no-sample-tables / enable_sample_tables=False) and points to the fix (re-prepare without the flag).
- SqliteIndexReader now runs a one-time check on first use of any samples-touching method. get_media_metadata is unaffected — the media_metadata table is independent.
- Test extended to cover the error type + message for three of the methods.

docs/source/basic/data_prep.md
```diff
@@ -686,6 +686,16 @@ The `media_metadata` table is used to store the media metadata for the selected
...
+
+For very large datasets (100M+ samples), populating the `samples` and `sample_parts` tables can dominate `energon prepare` runtime. If the dataset will only be consumed sequentially (not as an auxiliary dataset), you can skip these tables:
```
voegtlel: I think that this is not 100% correct: Datasets are not consumed sequentially for training. The only thing is that they cannot be "auxiliary" any more (and also not be mounted).

src/megatron/energon/flavors/webdataset/indexing.py
```diff
@@ -374,6 +413,25 @@ def __init__(self, sqlite_path: EPath):
...
+        """
+        if self._samples_table_checked:
+            return
+        assert self.db is not None, "Database is closed"
+        row = self.db.select_one(
+            "SELECT name FROM sqlite_master WHERE type='table' AND name='samples'"
+        )
+        if row is None:
+            raise MissingSamplesTableError(self.sqlite_path)
+        self._samples_table_checked = True
```
voegtlel: I think this needs 3 states? True / False / not checked yet? Otherwise this will just return even if it failed before (that's not a guaranteed failure). Can we instead implement this this similar to `db_has_sample_parts`? (i.e. evaluate in the `SqliteITarEntryReader`).

voegtlel (changes requested):

src/megatron/energon/flavors/webdataset/indexing.py
pei-li-hedgehog: `i.e. evaluate in the SqliteITarEntryReader` -> you mean the check should live in SqliteITarEntryReader, not spread across every method in SqliteIndexReader?

src/megatron/energon/flavors/webdataset/itar_reader.py
```diff
@@ -507,6 +511,10 @@ def __init__(
         self.sqlite_reader = SqliteIndexReader(sqlite_path)
 
         self.db_has_sample_parts = self.sqlite_reader.db_has_sample_parts()
+        self.db_has_samples = self.sqlite_reader.db_has_samples()
+
+        if not self.db_has_samples:
+            raise MissingSamplesTableError(sqlite_path)
```
voegtlel:
```suggestion
        if not self.sqlite_reader.db_has_samples():
            raise MissingSamplesTableError(sqlite_path)
```

No need to store to a property, if it'll raise for `False`.

src/megatron/energon/flavors/webdataset/itar_reader.py
```diff
@@ -484,7 +484,8 @@ class SqliteITarEntryReader(ITarReader[str]):
     """
 
     sqlite_reader: SqliteIndexReader
-    db_has_sample_parts: int
+    db_has_sample_parts: bool
+    db_has_samples: bool
```
voegtlel:
```suggestion
    db_has_sample_parts: bool
```

(see below)

voegtlel (changes requested):
Just this very small edit

pei-li-hedgehog:
Suggestions applied. Thx!

philipp-fischer:
I'd rather vote for improving the speed of creating those sqlite tables instead of introducing a flag to skip them. Skipping them means we will have datasets that are incompatible with some of energon's feature such as `energon mount`.

psycherun-kaiko:
How about doing both? I agree improving the indexing speed would be very helpful in general, but in some cases the sample-level indices might really be unnecessary. 

Moreover, this path is opt-in. We could additionally make the potential failures more clear to the user via docs, or failing fast. 

What do you think?

PR #235, opened 2026-05-20
Fix sample printer

voegtlel:
Indentation was broken. Print wasn't really well readable

### DIFF
--- a/src/megatron/energon/sample_utils.py
+++ b/src/megatron/energon/sample_utils.py
@@ -9,6 +9,7 @@ from typing import Any
 
 import numpy as np
 import torch
+import yaml
 
 
 def default_get_batch_keys(batch: Any) -> list[str] | None:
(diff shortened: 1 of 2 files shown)


philipp-fischer:
I don't really like the very custom branch-by-branch indentation algorithm. Also it would break for (nested) dataclasses. 
I brainstormed with codex and came up with a clean and simple approach. Curious what you think about it:

The idea is to first convert into a clean tree-like structure and then render as ordinary yaml.

I am pasting the rather short code here:
Step 1:
```
def to_printable(value):
    if isinstance(value, torch.Tensor):
        return summarize_tensor(value)

    if isinstance(value, np.ndarray):
        return summarize_array(value)

    if dataclasses.is_dataclass(value):
        return {
            type(value).__name__: {
                field.name: to_printable(getattr(value, field.name))
                for field in dataclasses.fields(value)
            }
        }

    if isinstance(value, dict):
        items = list(value.items())
        out = {k: to_printable(v) for k, v in items[:25]}
        if len(items) > 25:
            out["..."] = f"and {len(items) - 25} more items"
        return out

    if isinstance(value, (list, tuple)):
        out = [to_printable(v) for v in value[:10]]
        if len(value) > 10:
            out.append(f"... and {len(value) - 10} more items")
        return out

    if isinstance(value, str):
        return value[:1000] + "... (truncated)" if len(value) > 1000 else value

    if isinstance(value, (int, float, bool, type(None))):
        return value

    return repr(value)[:200]
```

Step 2 (render):
```
yaml.safe_dump(
    to_printable(sample),
    sort_keys=False,
    default_flow_style=False,
    width=120,
).rstrip()
```

where `summarize_tensor` etc. are similar to existing summarizing code for these classes.

PR #237, opened 2026-05-28
Fix gc frozen objects

voegtlel:
* Fixes a bug if compilation happened before creating workers, trying to free cuda objects
* Let the agent search torch for more such objects and verify that they cause the same issue, adding them to the list

Minimal repro script (also for future reference, if we find this again):
```py
import gc
import os
import torch

def make_object():
    # Here testing the torch.cuda.Event
    e = torch.cuda.Event()
    e.record()
    torch.cuda.synchronize()
    return e

obj = make_object()
x = [obj]
x.append(x)
del obj

if os.fork() == 0:
    x = None
    gc.collect()
    os._exit(0)

pid, status = os.wait()
print(pid, status)
```

### DIFF
--- a/src/megatron/energon/wrappers/gc_dataset.py
+++ b/src/megatron/energon/wrappers/gc_dataset.py
@@ -22,6 +22,23 @@ try:
 except ImportError:
     TorchDynamoOutputGraph = None
 
+try:
+    from torch.distributed.distributed_c10d import Work as TorchDistributedWork
+except ImportError:
+    TorchDistributedWork = None
+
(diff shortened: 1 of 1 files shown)


philipp-fischer:
Should we add more CUDA-owning classes to the list?
Codex found these in addition:
`torch.cuda.CUDAGraph`, `torch.cuda.Stream` / `torch.Stream`, and `torch.cuda.ExternalStream`

voegtlel:
I tested those objects as well, they did not produce the crash

PR #242, opened 2026-06-09
fix: call post_initialize() for dict path in load_dataset()

### LINKED ISSUE #241: [Bug] load_dataset() with dict input skips post_initialize(), causing AssertionError
### Problem

`load_dataset()` accepts a `dict` as the `path` argument (useful for embedding metadataset definitions inline in training configs), but the dict code path does not call `post_initialize()` on the resulting `Metadataset`. This leads to `AssertionError: assert self._dataset is not None` when the dataset is subsequently used (e.g., in `get_datasets()`).

The file-based code paths (`EnergonDatasetType.METADATASET` and `WEBDATASET`/`JSONL`) both call `post_initialize()` correctly.

### Reproduction

```python
from megatron.energon import get_train_dataset

metadataset_dict = {
    "__module__": "megatron.energon.metadataset.metadataset",
    "__class__": "Metadataset",
    "splits": {
        "train": {
            "datasets": [
                {"path": "/path/to/prepared/dataset", "weight": 1.0},
            ]
        }
    }
}

# This triggers the dict branch in load_dataset()
dataset = get_train_dataset(
    metadataset_dict,
    batch_size=2,
    task_encoder=my_task_encoder,
    worker_config=worker_config,
)
# → AssertionError in Metadataset.get_datasets(): assert self._dataset is not None
```

### Root Cause

In [`src/megatron/energon/metadataset/loader.py`](https://github.com/NVIDIA/Megatron-Energon/blob/main/src/megatron/energon/metadataset/loader.py#L25-L31):

```python
def load_dataset(path, **kwargs):
    if isinstance(path, dict):
        mds = load_config(
            path,
            default_type=Metadataset,
            default_kwargs=dict(path=EPath("/dict"), **kwargs),
        )
        return mds           # ← post_initialize() is NOT called

    # ... file path branch:
    mds = load_config(...)
    mds.post_initialize()    # ← called here
    return mds
```

The fix is a one-line addition (`mds.post_initialize()`) in the dict branch.

### Environment

- megatron-energon: installed from NGC NeMo 26.04 container
- Python 3.12

### Use Case

We embed metadataset definitions as dicts in OmegaConf training configs (passed through Megatron-Bridge's `EnergonProvider.path` field) to keep experiment configurations self-contained in a single YAML file. The dict path support exists in the code but is incomplete without `post_initialize()`.

nayopu:
Closes #241.

## Summary

`load_dataset()` accepts a `dict` as the `path` argument (for embedding metadataset definitions inline), but the dict code path does not call `post_initialize()` on the resulting `Metadataset`. This causes `AssertionError: assert self._dataset is not None` in `get_datasets()`.

The file-based code paths (`METADATASET`, `WEBDATASET`/`JSONL`) and `prepare_metadataset()` all call `post_initialize()` after `load_config()`. The dict branch is the only path that skips it.

## Changes

- Add `mds.post_initialize()` in the dict branch of `load_dataset()`, matching the existing file-based branches.

## Testing

Verified end-to-end with Megatron-Bridge `EnergonMultiModalDataModule` — passing a metadataset dict through `get_train_dataset()` now correctly initializes and returns data iterators without `AssertionError`.

### DIFF
--- a/src/megatron/energon/metadataset/loader.py
+++ b/src/megatron/energon/metadataset/loader.py
@@ -25,6 +25,7 @@ def load_dataset(
             default_type=Metadataset,
             default_kwargs=dict(path=EPath("/dict"), **kwargs),
         )
+        mds.post_initialize()
         return mds
     path = EPath(path)
     ds_type = get_dataset_type(path)


philipp-fischer:
Next time, please create PR to `develop`, not to `main`.

PR #249, opened 2026-07-02
Support GPU Image Decoding

Queuecumber:
This PR adds support for image decoding on GPU using NVImageCodec

NVImageCodec supports CUDA accelerated decoding of most common image formats and in particular supports the hardware JPEG decoding chip on modern NVIDIA GPUs. 

Users can now specify a `image_decode_device` argument which can be `cpu`, `gpu` or an integer CUDA device ordinal. This will transparently invoke NVImageCodec and return a tensor already on GPU. If NVImageCodec fails for any reason, the current behavior is to keep the webdataset autodecode path and fall back on it for CPU decoding.


Queuecumber:
## Notes
1. the GPU tests are all configured to skip if CUDA isn't available. I run them locally on a GPU, we might want to think about how best to test GPU-requiring code since it's about to get more relevant
2. Do we need to do anything w.r.t. free threaded python and torch interop to make this usable?
3. Currently the CUDA device selection for GPU decoding is *not* saved in `_config` is that OK?

Queuecumber:
Setting as draft pending nogil support

PR #250, opened 2026-07-10
Use NVIDIA sphinx theme

maanug-nv:
Update `pyproject.toml` and `conf.py` in docs to use the `nvidia-sphinx-theme` used by other Megatron Github repos' docs pages.


docs/source/conf.py
```diff
@@ -122,14 +122,20 @@
...
 #
-html_theme = "sphinx_rtd_theme"
+html_theme = "nvidia_sphinx_theme"
 
 # Theme options are theme-specific and customize the look and feel of a theme
 # further.  For a list of options available for each theme, see the
 # documentation.
 #
 html_theme_options = {
-    "prev_next_buttons_location": "both",
```
maanug-nv, line 132: seems the nvidia theme doesn't have support for this option.

docs/source/conf.py
```diff
@@ -69,11 +69,6 @@
 autodoc_typehints = "description"
 autodoc_typehints_format = "short"
 
-# Mock everything here, so that not just autodoc, but also sphinx_click can make use of the mock imports
```
maanug-nv, line 72: According to `sphinx-click` [documentation](https://sphinx-click.readthedocs.io/en/latest/usage/#mocking), it defaults to using `autodoc_mock_imports`.

Currently, these lines produce the following build error:
```
.../energon/nv-sphinx-theme/docs/source/api/cli.md:22: ERROR: Failed to import "main" from "megatron.energon.cli.main". The following exception was raised:
  File ".../energon/nv-sphinx-theme/.venv/lib/python3.12/site-packages/sphinx_click/ext.py", line 459, in _load_module
    with mock(self.env.config.sphinx_click_mock_imports):
         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
TypeError: 'module' object is not callable [docutils]
```

The reason for this error is that `sphinx.ext.autodoc.mock` is exported as a context manager, but the `from sphinx.ext.autodoc.mock` import is changing that imported path into the module, which is affecting later imports like the one in the error from `sphinx_click`.

maanug-nv: This became an issue with Sphinx 9.1.0 , and i think it's treated as a warning, so that's probably why it wasn't noticed.

voegtlel (approved):

voegtlel (changes requested):
Okay, I checked a local build. Everything looks good, except this: 
image
Any thought what's broken here?
(this is the automatic module documentation at http://127.0.0.1:8000/api/modules_data.html#megatron.energon.BatchDataset after running `just docs && just docs-serve`)

voegtlel:
> Okay, I checked a local build. Everything looks good, except this: ... Any thought what's broken here? (this is the automatic module documentation at http://127.0.0.1:8000/api/modules_data.html#megatron.energon.BatchDataset after running `just docs && just docs-serve`)

Just checked. We had some quirks for the old template, which are not needed with the nv template. Fixing those now, committing to your branch. Then this should be good 👍

voegtlel (approved):

PR #252, opened 2026-07-13
Support configurable multiprocessing start method (`forkserver`/`spawn`) to avoid fork deadlocks

Thaurun:
## Problem

The dataloader can **deadlock with the default `fork` start method** in environments where the
parent process already has other threads holding locks (OpenMP, MKL, CUDA, HuggingFace tokenizers,
logging handlers, etc.). `fork()` only copies the calling thread, so locks held elsewhere stay
permanently locked in the worker → deadlock.

## Solution

Make the multiprocessing start method configurable (`fork` / `forkserver` / `spawn`). `forkserver`/
`spawn` start a fresh interpreter and avoid inherited locks. Default stays `fork` — no behavior change.

## Changes

**Configurable start method**
- `loader.py`: `get_loader` / `get_savable_loader` accept `multiprocessing_context: str = "fork"`.
- `savable_loader.py`: loaders create queues + torch `DataLoader` context from
  `torch.multiprocessing.get_context(mp_ctx)`; `config()` now includes `multiprocessing_context`.
- `worker.py`: `WorkerConfig` gains `__getstate__`/`__setstate__` that drops the non-picklable
  `data_parallel_group` (ProcessGroup) on serialization.

**Make dataset callables picklable** (required because `forkserver`/`spawn` serialize the dataset to workers)
- `default_generic_webdataset.py`: lambdas → `functools.partial` of `_part_in_set`, `_apply_field_map`, `_wrap_sample`.
- `joined_webdataset.py`: joiner lambda → `functools.partial(_join_samples, ...)`.
- `crude.py`: default lambdas → module-level `_always_true_part` / `_identity_sample`.
- `mix_batch_dataset.py`: default `batch_mix_fn` → `_identity_batch_mix`.
- `iter_map_dataset.py`: default `len_map_fn` → `_identity_len`.

**Supporting**
- `base_dataset.py`: `_function_config` renders `functools.partial` as readable `functools.partial(<module>.<qualname>)`.
- `tests/test_forkserver.py` (new): save/restore + fork-order equivalence under `forkserver`.
- `tests/test_metadataset.py`: updated reference `config()` for the new `multiprocessing_context` key and `sample_loader` repr.

## Usage

```python
loader = get_savable_loader(dataset, multiprocessing_context="forkserver")  # or "spawn"

### DIFF
--- a/src/megatron/energon/flavors/webdataset/default_generic_webdataset.py
+++ b/src/megatron/energon/flavors/webdataset/default_generic_webdataset.py
@@ -2,6 +2,7 @@
 # SPDX-License-Identifier: BSD-3-Clause
 
 import dataclasses
+import functools
 from typing import Any, Callable, Dict, Generic, List, Optional, TypeVar, Union
 
 from megatron.energon.epathlib import EPath
@@ -13,6 +14,27 @@ from megatron.energon.flavors.webdataset.structs import FilteredSample
 from megatron.energon.module_loader import ModuleLoader
 
 T_sample = TypeVar("T_sample", covariant=True)
+
+
+def _part_in_set(part: str, parts) -> bool:
+    """Picklable replacement for ``lambda part: part in parts`` (forkserver/spawn support)."""
+    return part in parts
+
+
+def _apply_field_map(sample: Dict[str, Any], fields) -> Dict[str, Any]:
+    """Picklable replacement for the field_map sample loader lambda."""
+    return {k: field_access(sample, v) for k, v in fields.items()}
+
+
+def _wrap_sample(sample: Dict[str, Any], inner: Callable, subflavors) -> Dict[str, Any]:
+    """Picklable replacement for the sample wrapping lambda."""
+    return {
+        "__key__": sample["__key__"],
+        **inner(sample),
+        "__restore_key__": sample["__restore_key__"],
+        "__subflavors__": subflavors,
+        "__sources__": sample["__sources__"],
+    }
 
 
 class DefaultGenericWebdatasetFactory(BaseWebdatasetFactory[T_sample], Generic[T_sample]):
@@ -61,7 +83,7 @@ class DefaultGenericWebdatasetFactory(BaseWebdatasetFactory[T_sample], Generic[T
                 sample_loader = sample_loader
             if isinstance(part_filter, list):
                 parts = set(part_filter)
-                part_filter = lambda part: part in parts
+                part_filter = functools.partial(_part_in_set, parts=parts)
             elif isinstance(part_filter, str):
                 part_filter = module_loader.get_function(
                     part_filter, "part_filter", relative_path=path / MAIN_FOLDER_NAME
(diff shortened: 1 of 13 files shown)

### CODE REVIEW

File: src/megatron/energon/flavors/webdataset/default_generic_webdataset.py [hunk 1]
```diff
@@ -84,19 +106,15 @@ def __init__(
...
-            "__key__": sample["__key__"],
-            **inner_sample_loader(sample),
-            "__restore_key__": sample["__restore_key__"],
-            "__subflavors__": self.subflavors,
-            "__sources__": sample["__sources__"],
-        }
+        self._sample_loader = functools.partial(
+            _wrap_sample,
+            inner=inner_sample_loader,
+            subflavors=subflavors or {},
```
VVsssssk: Does capturing subflavors=subflavors or {} here pose a behavior regression risk?

If dataset.yaml doesn't have subflavors, but the caller/Metadataset subsequently executes dataset.subflavors.update(subflavors) at src/megatron/energon/dataset_config.py:138, the __subflavors__ in the samples won't see the update.

src/megatron/energon/worker.py
```diff
@@ -262,6 +262,24 @@ def config(self) -> Dict[str, Any]:
...
+        """Return a picklable copy of the state.
+
+        ``data_parallel_group`` holds a ``torch.distributed.ProcessGroup`` which is not
+        picklable. It is only needed in the main process (e.g. for ``global_rank()``), so it
+        is dropped here. This is required when the dataloader uses a multiprocessing start
+        method other than ``fork`` (e.g. ``forkserver``/``spawn``), where the dataset is
+        serialized to the worker processes instead of inherited via fork.
+        """
+        state = {name: getattr(self, name) for name in type(self).__slots__}
+        state.pop("data_parallel_group", None)
```
VVsssssk, line 279: Only data_parallel_group was removed — should _worker_debug_file also be removed? It may not be picklable either.

When worker_log_level >= 1 and worker_debug_path is set, the loader initialization opens the debug file in the main process; the dataset serialization may then fail when forkserver/spawn serializes it.

src/megatron/energon/flavors/webdataset/default_generic_webdataset.py
Thaurun: Good catch — yes, that was a regression. When subflavors is None, the two subflavors or {} expressions created distinct empty dicts, so later dataset.subflavors.update(...) no longer affected __subflavors__ on samples. Fixed by resolving the dict once and sharing that same object between self.subflavors and the picklable _wrap_sample partial.

src/megatron/energon/worker.py
Thaurun: Agreed. With worker_log_level >= 1 and worker_debug_path set, SavableDataLoader.__init__ already opens _worker_debug_file on the shared WorkerConfig before workers are spawned, so forkserver/spawn would try to pickle an open file handle. Fixed by also dropping _worker_debug_file (and _worker_debug_file_worker_id) in __getstate__; workers reopen the file lazily in worker_log.

Thaurun:
Addressed all review comments — fixed the shared subflavors dict, dropped _worker_debug_file from pickle state, and made FileStoreCachePool picklable for forkserver/spawn.

PR #264, opened 2026-09-05
introduce torch safe globals

dimapihtar:
Introduces `safe_globals` list for `torch.load()` to avoid `weights_only` usage which is very unsafe.

### DIFF
--- a/src/megatron/energon/safe_globals.py
+++ b/src/megatron/energon/safe_globals.py
@@ -1,3 +1,6 @@
+# Copyright (c) 2026, NVIDIA CORPORATION.
+# SPDX-License-Identifier: BSD-3-Clause
+
 import torch
 
 from megatron.energon.rng import SystemRngState
--- a/src/megatron/energon/safe_globals.py
+++ b/src/megatron/energon/safe_globals.py
@@ -6,6 +6,7 @@ import io
 import numpy
 import torch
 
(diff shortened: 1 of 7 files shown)


src/megatron/energon/safe_globals.py
```diff
@@ -0,0 +1,40 @@
...
+from megatron.energon.flavors.webdataset.sample_loader import SliceState
+from megatron.energon.rng import SystemRngState
+from megatron.energon.savable_loader import (
+    SavableDataLoaderState,
+    SavableDatasetCheckpoint,
+    SavableDatasetState,
+)
+from megatron.energon.state import FlexState
+
+SAFE_GLOBALS = [
```
philipp-fischer, line 18: Can we be sure the list is complete? 
Could there be cases where other types are (legally) contained in a checkpoint?

philipp-fischer (changes requested):
I had another look and I think we really have a problem with completeness of this class list.

Look at GroupBatchDataset for example with is used for [grouping](https://nvidia.github.io/Megatron-Energon/advanced/grouping.html). The user may use any hashable object as the grouping key, like an enum for example.

The keys will end up in the checkpoint and then an error will be raised upon loading the checkpoint, because we can never cover all hashable objects. Not sure how we can solve this.

PR #265, opened 2026-09-05
introduce SafeUnpickler

dimapihtar:
Introduces `SafeUnpickler` class with defined safe set of classes can be used with pickle load which substitutes unsafe usage of  `pickle.load()`.

### DIFF
--- a/src/megatron/energon/safe_globals.py
+++ b/src/megatron/energon/safe_globals.py
@@ -1,0 +1,34 @@
+# Copyright (c) 2025, NVIDIA CORPORATION.
+# SPDX-License-Identifier: BSD-3-Clause
+
+import numpy
+import pickle
+from unittest.mock import patch
+
+
+def _safe_pickle_load(file, **kwargs):
+    """Safe version of `pickle.load`."""
+    return SafeUnpickler(file, **kwargs).load()
+
+
+def safe_numpy_load(path, **kwargs):
+    """Safe version of `numpy.load` which calls `pickle.load`."""
+    with _pickle_patch_lock:
+        with patch('pickle.load', _safe_pickle_load):
+            return numpy.load(path, **kwargs)
+
+
+class SafeUnpickler(pickle.Unpickler):
+    """ """
+
+    _SAFE_CLASSES: frozenset = frozenset(
+        {
+        }
+    )
+
+    def find_class(self, module: str, name: str):
+        if (module, name) not in self._SAFE_CLASSES:
+            raise pickle.UnpicklingError(
+                f"Refusing to unpickle disallowed class '{module}.{name}' "
+            )
+        return super().find_class(module, name)
--- a/src/megatron/energon/safe_globals.py
+++ b/src/megatron/energon/safe_globals.py
@@ -1,9 +1,15 @@
 # Copyright (c) 2025, NVIDIA CORPORATION.
 # SPDX-License-Identifier: BSD-3-Clause
 
-import numpy
 import pickle
+import threading
 from unittest.mock import patch
+
+import numpy
+
+# Guards `pickle.load` patching in `safe_numpy_load`, so concurrent calls from multiple
+# threads don't race on the monkeypatch.
+_pickle_patch_lock = threading.Lock()
 
 
 def _safe_pickle_load(file, **kwargs):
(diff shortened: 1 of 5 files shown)


src/megatron/energon/safe_unpickler.py
```diff
@@ -0,0 +1,48 @@
...
+
+
+def safe_numpy_load(path, **kwargs):
+    """Safe version of `numpy.load` which calls `pickle.load`."""
+    with _pickle_patch_lock:
+        with patch('pickle.load', _safe_pickle_load):
+            return numpy.load(path, **kwargs)
+
+
+class SafeUnpickler(pickle.Unpickler):
```
philipp-fischer, line 7: Should the python module be called safe_unpickler.py instead of safe_globals maybe?

src/megatron/energon/safe_globals.py
```diff
@@ -0,0 +1,48 @@
...
+# threads don't race on the monkeypatch.
+_pickle_patch_lock = threading.Lock()
+
+
+def _safe_pickle_load(file, **kwargs):
+    """Safe version of `pickle.load`."""
+    return SafeUnpickler(file, **kwargs).load()
+
+
+def safe_numpy_load(path, **kwargs):
```
philipp-fischer: Is this being used anywhere? And if yes, loading npz might circumvent the guards.

src/megatron/energon/safe_globals.py
```diff
@@ -0,0 +1,48 @@
...
+from unittest.mock import patch
+
+import numpy
+
+# Guards `pickle.load` patching in `safe_numpy_load`, so concurrent calls from multiple
+# threads don't race on the monkeypatch.
+_pickle_patch_lock = threading.Lock()
+
+
+def _safe_pickle_load(file, **kwargs):
```
philipp-fischer: Is this being used anywhere?

philipp-fischer (changes requested):

src/megatron/energon/safe_unpickler.py
dimapihtar: done

src/megatron/energon/safe_globals.py
dimapihtar: removed it. Looks like energon doesn't call `numpy.load()` at all.

src/megatron/energon/safe_globals.py
dimapihtar: removed it. Looks like energon doesn't call `numpy.load()` at all.

tests/test_safe_unpcikler.py
```diff
@@ -0,0 +1,32 @@
...
+
+class UnsafeClass:
+    def __init__(self, value):
+        self.value = value
+
+    def __repr__(self):
+        return f"UnsafeClass(value={self.value})"
+
+
+class TestSafeUnpickler:
```
philipp-fischer, line 20: The test will not be executed by CI. It must derive from unittest like the other test cases.

src/megatron/energon/safe_unpickler.py
```diff
@@ -0,0 +1,28 @@
...
+    """A `pickle.Unpickler` that only constructs classes/functions explicitly allowlisted
+    in `_SAFE_CLASSES`, to protect against arbitrary code execution when unpickling data
+    from disk. Use `add_safe_classes` to extend the allowlist for custom payload types.
+
+    Usage: `SafeUnpickler(file).load()` instead of `pickle.load(file)`.
+    """
+
+    _SAFE_CLASSES: set = {
```
philipp-fischer, line 15: I think we have the same completeness issue as with the other PR. The file cache pool is meant to pickle and unpickle user data. That can literally be any data and it's had to come up with an exhaustive list.
A simple `torch.Tensor` for example would not be covered as of now, or the AVDecoder and AVData objects, we often use. I think there can be many more.

philipp-fischer (changes requested):
After thinking about this again: The file cache pool does not load externally provided pickle files but only files it generated itself. So is there even a security risk?

src/megatron/energon/safe_unpickler.py
dimapihtar: @philipp-fischer simple objects like `torch.Tensor` are covered by torch automatically.

you can see list of default safe globals defined by torch by running:
```
from torch._weights_only_unpickler import _get_allowed_globals
print(sorted(_get_allowed_globals()))
```

